PR #14, opened 2018-10-01
Gaussian Mixture Model

trappmartin:
Gaussian mixture model example.


cpfiffer:
Looks good to me, I have no problem with merging this over. I have some minor cosmetic changes I'd like to make, but those are exceptionally minor. Do you know if there's a quick way to make the labels larger on the plots? I've noticed that the text seems to be remarkably small when I build my plots.

trappmartin:
Maybe something like this: https://github.com/JuliaPlots/Plots.jl/issues/897 ? I'm not aware of a quicker way. You can set a theme using e.g. `theme(:ggplot2) ` but this doesn't seem to change the font size.

yebai:
Perhaps consider renaming the file to `1_GaussianMixtureModel` to keep naming consistency with existing tutorial?

trappmartin:
Thx, I'll add your suggestions asap. A reference to Bishop is a really good idea!

cpfiffer:
I'm going to merge this over and make a couple changes on the master branch, and try to get it up on the site sometime tonight.

PR #15, opened 2018-10-01
Bayesian HMM Tutorial

yebai:


cpfiffer:
With the most recent batch of commits, this one is close to ready. Would anyone mind taking a look at this and verifying that my models look reasonable?

markdown/4_BayesHmm.md
```diff
@@ -0,0 +1,310 @@
...
+Ns = 1:300
+
+# Make an animation.
+animation = @animate for (i, N) in enumerate(Ns)
+    m = m_set[N]; s = s_set[N];
+    p = plot(y, c = :red,
+        size = (500, 250), 
+        xlabel = "Time", 
+        ylabel = "State", 
+        legend = legend = :topright, label = "True data",
```
tobanw:
```suggestion
         legend = :topright, label = "True data",
```

markdown/4_BayesHmm.md
```diff
@@ -0,0 +1,310 @@
...
+        xlim = (0,15),
+        ylim = (-1,5)); 
+    plot!(p, m[s], c = :blue, label = "Sample $$N")
```
tobanw:
```suggestion
    plot!(p, m[s], c = :blue, label = "Sample $N")
```

markdown/4_BayesHmm.md
```diff
@@ -0,0 +1,310 @@
...
+
+
+Looks like our model did a pretty good job!
+
+
+## Detecting A Fair Coin
+
+Now it's all well and good to plot a line on a nice, clean dataset. But what if we are trying to detect when someone is cheating us in a coin flipping game? 
+
+Let's say we have a friend named Jane. Every once in a while we play a game with Jane where she flips a coin. Jane wins a dollar from us if it's heads, we win a dollar from Jane if it's tails. The problem is that Jane is sometimes a cheater — 20% of the time, she'll bring out a coin that comes up heads 90% of the time, and she decides to switch back to the fair coin 80% of the time so we don't become suspicious. Hidden Markov models are ideal for learning this kind of transition matrix, so let's get to it.
```
tobanw:
```suggestion
Let's say we have a friend named Jane. Every once in a while we play a game with Jane where she flips a coin. Jane wins a dollar from us if it's heads, we win a dollar from Jane if it's tails. The problem is that Jane is sometimes a cheater — 10% of the time, she'll bring out a coin that comes up heads 90% of the time, and she decides to switch back to the fair coin 80% of the time so we don't become suspicious. Hidden Markov models are ideal for learning this kind of transition matrix, so let's get to it.
```

markdown/4_BayesHmm.md
```diff
@@ -0,0 +1,310 @@
...
+````
+
+
+
+
+Time to declare our Turing model. 
+
+You will note that this model is very similar to the one we used in the previous example, but now the emission parameter `m` has two separate `Beta` priors, one showing our belief that there is a fair coin and an unfair coin. This will allow us to model the probability of getting a heads in each state, and establishing them as distinct priors helps to avoid the label-switching problem. 
+
+Additionally, when we observe each data point `y[i]`, we are using the `Bernoulli` distribution, which models this coin-flipping game closer than does the `Normal` dsitribution we used previously.
```
tobanw:
```suggestion
Additionally, when we observe each data point `y[i]`, we are using the `Bernoulli` distribution, which models this coin-flipping game closer than does the `Normal` distribution we used previously.
```

markdown/4_BayesHmm.md
```diff
@@ -0,0 +1,310 @@
...
+
+````julia
+# Turing model definition.
+@model BayesHmmCoins(y, K) = begin
+    num_obs = length(y)
+    
+    # State sequence.
+    s = tzeros(Int, num_obs)
```
tobanw:
```suggestion
    s = Vector{Integer}(undef, num_obs)
```

markdown/4_BayesHmm.md
```diff
@@ -0,0 +1,310 @@
...
+2. The emission matrix, which describes a typical value emitted by some state. In the plot above, the emission parameter for state one is simply one.
+3. The state sequence is our understanding of what state we were actually in when we observed some data. This is very important in more sophisticated HMM models.
+
+With this in mind, let's set up our model.
+
+````julia
+# Turing model definition.
+@model BayesHmm(y) = begin
+    # State sequence.
+    s = tzeros(Int, N)
```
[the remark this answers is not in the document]
cpfiffer: `tzeros(Int, N)` creates a `TArray{Int}` of length `N`. The `TArray` is required here because we are using the `PG` sampler. 

There's some more info on that [here](http://turing.ml/docs/guide/#working-with-libtaskjl).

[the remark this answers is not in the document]
tobanw: Good to know -- I was wondering what that was for. Thanks for the pointer.

tobanw:
I'd suggest adding some plots of the traces, and some discussion of convergence / burn-in. For example, this is how I plotted the diagonal elements of the transition matrix samples (i.e., state persistence probabilities):

```julia
T_diag_trace = [t[i][i] for t in c[:T], i in 1:K];

plot(T_diag_trace, ylim = (0,1),
     label = ["T[$i,$i]" for i = 1:K],
     xlabel = "Sample", ylabel = "Persistence probability"
)
```

cpfiffer:
@tobanw Would it be alright if I used that plotting code in the tutorial? It's very succinct.

tobanw:
@cpfiffer absolutely, please do.

PR #23, opened 2019-03-05
0_Introduction: Use gif instead of animation

ayan-b:
If I use `@animation`, I get some 100 pngs instead of a single animation, whereas `@gif` gives a nice animation. However, I am using `Julia v1.0.3` instead of `Julia v1.0.0` in which the notebook was written.


cpfiffer:
Thanks for adding that, I appreciate it!

Issue #1, opened 2018-09-06
Tutorials Roadmap


cpfiffer:
My understanding of this repo is that it is to be the home for all of Turing's educational materials. I've reviewed a lot of the existing stuff we have in TuringDemos and TuringExamples, but so far there's not a lot outside of brief code blurbs and a couple of Jupyter notebooks without a lot of commentary.

@trappmartin noted that [Edward's tutorials](http://edwardlib.org/tutorials/) are something of a target for us. I can see why, they are exceptionally thorough and very pleasant to read with a conversational approach. They also tend assume limited familiarity with probabilistic methods, which I think of as a plus.

So, here are my guidelines for what I think would make a good set of tutorials.

1. Highlight the ease of use of Turing. Easy to read, easy to write, easy to understand.
2. Usable by reasonably informed layman. Someone with a basic to intermediate understanding of statistics should be able to understand most of the tutorial and expand it to their problem. This doesn't necessarily mean that we want to dumb everything down -- I still think there's room to showcase some really complex and nifty stuff, but we should try and make everything as accessible as is reasonable.
3. Solve an example problem with real data. This might be any of the data sets in [RDatasets](https://github.com/johnmyleswhite/RDatasets.jl) or [MLDatasets](https://github.com/JuliaML/MLDatasets.jl). In cases where a real-world data set might not be available or isn't applicable (are there any of those in this field? I have no idea) we should make sure that synthetic data is described (table headers, a chart or two, etc) so people can easily see what it looks like.
4. Consistent in tone, information content, and pedagogical approach.

I'll be adding issues for each tutorial type we should eventually have, along with what we have in place elsewhere to see if we can adapt some of the models already written. More to come on that, but figured I should get out what kinds of things I think are valuable in a tutorial.

trappmartin:
Regarding 3., I think it should be possible to find a dataset in the mentioned pkgs or via other sources for every show-case application. Synthetic data, however, might sometimes simplify the documentation, e.g. for a GMM example.

mkborregaard:
Hey, I just found this repo. The first notebook "introduction" and the planned expansions look really sweet. Looking forward to following this!

cpfiffer:
@mkborregaard Happy we could showcase some of our cool stuff. Let us know if there's anything specific you think might be interesting to include in the tutorials.

mkborregaard:
I will! By the way, StatPlots is better at plotting distributions than just Plots - if you use StatPlots you can replace the line 
[...]

PR #25, opened 2019-03-10
Added tutorial for Poisson Regression

saumyagshah:
Added a tutorial for implementing Poisson Regression in Turing. This was ported from one of the PyMC3 [example notebooks](https://docs.pymc.io/notebooks/GLM-poisson-regression.html). I tried to stick to the Tutorials Roadmap described in #1 .


PoissonRegression.ipynb
```diff
@@ -0,0 +1,5371 @@
+{
```
cpfiffer: <p>&gt; on a given day.This example describes a method </p><p><br></p><p>The above needs a space between the period and "This".</p><p><br></p><p><br></p>

 _Reply via <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/25/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

cpfiffer: <ul><li>"Medicines" should just be "Medicine". </li><li>Printing the first and last 10 lines is a lot. Print just the first five rows, or, if you want to show people a mix of the data, show five random rows from the dataframe.</li></ul>

 _Reply via <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/25/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

cpfiffer: <ul><li>Lines 1-2 in this block: `took a medicine` should be `took medicine`</li><li>As above, printing 10 lines at the top and bottom of the dataframe is a lot. Consider printing just five total rows.</li><li>I would lower your sample size from `q=1000` to `q=100` to make sampling much quicker. Currently it takes a ridiculous amount of time to sample a single chain, and people may give up if they're following this tutorial and see that they have to wait 750 seconds per chain.</li></ul>

 _Reply via <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/25/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

cpfiffer: <ul><li>`his medicines` should be `his medicine` everywhere in this paragraph.</li></ul><p><br></p><p><br></p>

 _Reply via <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/25/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

cpfiffer: <p>Indexing the different categories of people like this (`df[1:1000,1]`) is pretty clunky. I would recommend sorting by indicator variable (whether they had a 0 for alcohol and meds, etc).</p><p><br></p><p>You are also contaminating your samples here: `df[1:1000,1]` and `df[1000:2000,1]` both share sample `1000`. It's the same for the others as well.</p>

 _Reply via <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/25/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

cpfiffer: <ul><li>`no medicines` should be `no medicine`</li></ul><p><br></p>

 _Reply via <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/25/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

cpfiffer: <p>I would remove the `y = tzeros(Int, n)` line, it's likely to cause confusion.</p>

 _Reply via <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/25/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

cpfiffer: <p>Add a semicolon at the end of the `sample` call (`chain = mapreduce(. . .);`) so the chain description doesn't print at the end of the output log. You should add some text in a new cell below this saying </p><p><br></p><p>"We can take a look at our parameter estimates using `describe(chain)`:</p><p><br></p><p>And then include a code cell where you call `describe(chain)`. It's important to separate out sampling and looking at the chain so the whole document is easier to read.</p>

 _Reply via <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/25/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

cpfiffer: <p>Don't just launch into running the diagnostics. Write some text talking about what you are about to do and what you're hoping to find out.  What is a problem if this test fails?</p>

 _Reply via <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/25/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

cpfiffer: <p>I recommended calling `describe(chain)` above, so you can remove all of these `print` calls and manually calculating the means of each parameter.</p><p><br></p><p>You can also simply call `plot(chains)` instead of plotting `plot(chain)`, MCMCChains is pretty good about plotting multiple chains together.</p><p><br></p>

 _Reply via <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/25/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

cpfiffer: <p>&gt; The exponentiated mean of the coefficient <code>b1</code> is roughly half of that of <code>b2</code> which makes sense because in the data that we generated, the number of sneezes was more sensitive to the medicinal intake as compared to the alcohol consumption</p><p><br></p><p>This is a really long sentence. I would add a period at the end of `b2`, and then start the following sentence with "This makes sense because...".</p>

 _Reply via <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/25/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

cpfiffer: <p>There should be only one output per code block -- either print the variances or plot the figure, but not both. These need to be broken into two code blocks. </p><p><br></p><p>You might also consider just running `describe(chains[201:end, :, :])`, which presents standard deviation. </p>

 _Reply via <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/25/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

cpfiffer:
I've added some comments using some weird service NBReviewer because it is a huge pain to review Jupyter notebooks. My comments came out a bit garbled, so let me know if everything makes sense.

PoissonRegression.ipynb
saumyagshah: <p>Added the period</p>

saumyagshah: <p>Changed "medicines" to "medicine"</p><p>Printed five random rows from the dataframe</p>

saumyagshah: <p>Made the changes</p>

saumyagshah: <p>Changed. Also included a period at the end of the last sentence.</p>

saumyagshah: <p>Changed. Used indicator variable to get the data for plotting. For the first line, I used:</p><p>p1 = Plots.histogram(df[(df[:alcohol_taken] .== 0) .&amp; (df[:nomeds_taken] .== 0), 1], title = "no_alcohol+meds")&nbsp;&nbsp;</p><p><br></p>

saumyagshah: <p>Changed.</p>

saumyagshah: <p>Forgot to remove the comment. Corrected the error.</p>

saumyagshah: <p>Added semicolon, the text and the code for describe(chains) as you mentioned. Looks neater than before.</p>

saumyagshah: <p>Changed. Now, this block includes just one line: `plot(chains)`.</p>

saumyagshah: <p>I think that the print statements should still be included because the `describe(chains)` function gives the means. What we require is the exponentiated means to give a better sense of the relative values.</p>

saumyagshah: <p>Changed.</p>

saumyagshah: <p>Separated the two code blocks. Replaced the printing part by describe(chains[401:end,:,:]). I used 401 instead of 201 as I found that one of the chains was taking a bit longer to converge.</p>

saumyagshah: <p>Added the following lines explaining the reason behind using the diagnostic, what we expect to find out and how to proceed if the test fails:</p><p><br></p><p>We use the Gelman, Rubin, and Brooks Diagnostic to check whether our chains have converged. Note that we require multiple chains to use this diagnostic which analyses the difference between these multiple chains.&nbsp;</p><p><br></p><p>We expect the chains to have converged. This is because we have taken sufficient number of iterations (1500) for the NUTS sampler. However, in case the test fails, then we will have to take a larger number of iterations, resulting in longer computation time.</p>

saumyagshah: I just noticed that taking due to `q = 100`, the chains did not converge. Should we use a bigger value for `q`

saumyagshah: As I mentioned in one of the earlier comments (increasing the value of `q`), I noticed that the chains did not converge.

cpfiffer: It might be worth sampling a longer chain instead of using a really big sample -- try running the sampler for more iterations and then bringing `q` back up if that doesn't work.

cpfiffer: Good addition!

cpfiffer: Ah, okay. Then cut `describe` entirely, and make sure the `print` calls are in a separate block from `plot`. I'd also add a line or two in the text block above describing why you're exponentiating the means.

cpfiffer: I wouldn't --- use the whole chain starting from 201. Usually we only cut out at 201 to remove the adaptation samples. It seems a little disingenuous to remove an extra 200 samples because the chain lagged a little bit.

saumyagshah: Okay, reverted back to cutting out at 201 samples.

saumyagshah: Used print calls instead of `describe(chains)`. Added a line explaining the reason behind using exponentiated means:
We take the exponent of the means to get a better comparison of the relative values of the coefficients.

saumyagshah: I tried using `q = 100` and 2000 iterations but it did not converge. Using `q = 100` and 2500 iterations worked well and the chains converged. For each chain, the time for each chain was about 110-140 seconds. Do you think this is reasonable?

cpfiffer: Yeah, and it's much less time than using `q=1000`. That's about par for a model of this type.

PR #27, opened 2019-03-18
typo in comment corrected

ritesh99rakesh:
The comment mentioned that we taking 100 data points but we had N=30, so changed it to 30 data points.


cpfiffer:
Thanks!

PR #30, opened 2019-03-24
Negative Binomial Regression Implementation

ritesh99rakesh:
This is a tutorial for Negative Binomial Regression using Turing. All code is commented and theory is also explained.


ritesh99rakesh:
@cpfiffer I am getting `NaN` values while sampling. Could you please explain to me how I can correct it. Regards.

PR #31, opened 2019-03-24
Negative Binomial Regression Implementation

ritesh99rakesh:
This is a tutorial for Negative Binomial Regression using Turing. All code is commented and theory is also explained.


ritesh99rakesh:
@cpfiffer I am getting NaN values while sampling. Could you please explain to me how I can correct it. Regards.

cpfiffer:
If you are talking about this:

```
┌ Info: [Turing] looking for good initial eps...
└ @ Turing.Inference /home/riteshk/.julia/packages/Turing/yF1Mq/src/inference/support/hmc_core.jl:245
┌ Warning: Numerical error in gradients. Rejecting current proposal...
└ @ Turing.Core /home/riteshk/.julia/packages/Turing/yF1Mq/src/core/ad.jl:169
┌ Warning: grad = Real[NaN, NaN, NaN, NaN]
└ @ Turing.Core /home/riteshk/.julia/packages/Turing/yF1Mq/src/core/ad.jl:170
┌ Info: [Turing] found initial ϵ: 0.013036584854125979
└ @ Turing.Inference /home/riteshk/.julia/packages/Turing/yF1Mq/src/inference/support/hmc_core.jl:237
┌ Warning: Numerical error in gradients. Rejecting current proposal...
└ @ Turing.Core /home/riteshk/.julia/packages/Turing/yF1Mq/src/core/ad.jl:169
┌ Warning: grad = Real[NaN, NaN, NaN, NaN]
└ @ Turing.Core /home/riteshk/.julia/packages/Turing/yF1Mq/src/core/ad.jl:170
┌ Info:  Adapted ϵ = 0.010067677574384128, std = [1.0, 1.0, 1.0, 1.0]; 200 iterations is used for adaption.
```

Then that is normal. You don't need to do anything there.

I believe I mentioned this in Slack, but this PR is remarkably similar to an existing PR (#25) --- what is this tutorial adding that's not found there?

ritesh99rakesh:
@cpfiffer This notebook demos negative binomial regression using Turing. It closely follows #25  Poisson regression example except for the data here is negative binomially distributed instead of Poisson distributed.

Negative binomial regression is used to model count data for which the variance is higher than the mean. The negative binomial distribution can be thought of as a Poisson distribution whose rate parameter is gamma distributed, so that rate parameter can be adjusted to account for the increased variance.
I have tried to show the same principle, in the notebook, where I have shown data derived from Poisson and NB Regression and compared the two. I attempted to implement one example from PyMC3.
Please let me know your views on it. And if there is anything I can do to improve it.

cpfiffer:
Could you: 

1. Repackage this in a `negativebinomial.jl` file with minimal comments
2. Keep all the setup code and the model, and remove anything that has to do with plotting, diagnostics, or otherwise describing the chain
3. Submit a PR to this repo instead: [https://github.com/TuringLang/TuringExamples](https://github.com/TuringLang/TuringExamples)

We're going to try to get smaller tutorials like this into our model zoo, so that people can browse through them and figure out how to make models a little quicker. I'

ritesh99rakesh:
@cpfiffer I have added the `negativebinomial.jl` file here. Please review it. If it looks good then I will open a PR on  https://github.com/TuringLang/TuringExamples

cpfiffer:
That is perfect, thank you!

PR #32, opened 2019-03-24
Marginalized Gaussian Mixture Model Implementation

ritesh99rakesh:
This is a tutorial for the Marginalized Gaussian Mixture Model using Turing. All code is commented and theory is also explained.
Ported from: https://docs.pymc.io/notebooks/marginalized_gaussian_mixture_model.html


ritesh99rakesh:
I am getting error related to issue: TuringLang/Turing.jl#722
@cpfiffer, we discussed regarding this.
Once it is correct, this tutorial should work.

PR #38, opened 2019-03-30
Added Model Comparison Tutorial

saumyagshah:
Added tutorial for Model Comparison. Ported from [this](https://docs.pymc.io/notebooks/model_comparison.html) PyMC3 noteboook.


6_Model Comparison.ipynb
```diff
@@ -0,0 +1,6627 @@
+{
```
cpfiffer: <ul><li>`example notebook of PyMC3` =&gt; `example notebook from PyMC3`</li><li>I would add a little more discussion here about the general difficulties involved in model selection, particularly what is meant by "best" and "goodness". </li><li>`We try out two models` =&gt; `We try two models`</li></ul>

 _Reply via <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/38/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

cpfiffer: <p>Add a ; at the end of the turnprogress line so it doesn't print out "false" at the end.</p>

 _Reply via <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/38/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

cpfiffer: <p>I would say "standard deviation" instead of "error", because stats people might assume you're talking about an error term and not precision about a distribution's mean.</p>

 _Reply via <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/38/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

cpfiffer: <p>Add a semicolon at the end of the sigma definition so it doesn't print out the array you defined.</p>

 _Reply via <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/38/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

cpfiffer: <p>I would add a comment here as to why it is you (and the PyMC3 people) are using 10^6 as a prior on precision.</p>

 _Reply via <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/38/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

cpfiffer: <p>This flexibility provided =&gt; The flexibility provided</p>

 _Reply via <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/38/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

cpfiffer: <ul><li>Add periods for each sentence at the end of your bullet points.</li><li>I don't really know what "This is the required posterior" means -- elaborate a bit more?</li></ul>

 _Reply via <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/38/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

cpfiffer: <ul><li>"Now, we taka a hierarchical model." Should probably be something like "Now we will construct a hierarchical model."</li></ul>

 _Reply via <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/38/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

cpfiffer: <p>Make the header here something like "Model Comparison with WAIC" so it's  slightly more clear that it relates to model comparison.</p><p><br></p><p>Also, could you add a note that this step is likely to be temporary until MCMCChains supports these functions natively?</p>

 _Reply via <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/38/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

cpfiffer: <p>Print out the values here. People shouldn't have to scroll up to see what the `p_waic` values were for each model.</p><p><br></p><blockquote>However, when it comes to performance,</blockquote><p>What do you mean by performance? You mention it here, but someone may not know to what you are referring. </p>

 _Reply via <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/38/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

saumyagshah: - Changed

- I added the following paragraph:
By the 'goodness' of a model, we are referring to how compatible it is to the data. For example, if we have a linear and a quadratic model, we ask the question: Which model best fits the data? Note that here, we cannot use the likelihood to compare the two models. The likelihood simply increases as you increase the complexity of the model. So, we need to find a technique that compares the two models while also accounting for overfitting when using complex models.

- Changed

saumyagshah: Added ;

saumyagshah: Changed

saumyagshah: Done.

saumyagshah: Changed.

saumyagshah: - Added periods.
- Replaced "This is the required posterior" with "These sampled values are an approximation of our model's posterior."

saumyagshah: Replaced "Now, we taka a hierarchical model." with "Now, we will construct a hierarchical model."

saumyagshah: Changed the header.

Added these lines: "The code block below manually calculates the WAIC. This is likely to be temporary until `MCMCChains` supports these functions natively."

saumyagshah: Made some changes. I replaced the mention of 'perform' everywhere with 'fit', 'goodness' and 'compatible'. These words were used to explain the concept of model comparison at the beginning.

It now looks like:
# Interpretation
Since the hierarchical model seems to be more complex and flexible, one might expect it to fit better on our data. It does have higher complexity which can be seen by comparing the p_waic values for both the models.

Block for printing the p_pooled, p_hierarchical values.

So, we get:
`p_hierarchical` > `p_pooled`
However, when it comes to deciding which model better fits the data ('goodness' of the model, as discussed earlier), we can see that the computed `waic` values for the models are approximately equal.

Block for printing the p_pooled, p_hierarchical values.

This means that both models are equally compatible with our data. Hence, we may be tempted to use the more complex model but this example shows that the simpler pooled model compares equally well.

saumyagshah: Added the following explanation:
 "We take an uninformative prior with a high standard deviation, which is close to a uniform distribution. This effectively means that we are taking no assumption on the parametrization."

PR #43, opened 2019-04-08
Ported Probabilistic PCA Tutorial from Edward

sshkhr:
I have tried to port [Edward's tutorial on Porbabilistic PCA](http://edwardlib.org/tutorials/probabilistic-pca) as closely as possible to Turing. While running HMC I found that it was extremely slow with 5000 data points and hence I've run on only 50 data points 

Please give feedback in order to improve this or if it looks okay I can add the markdown files for the website as well. Thanks


cpfiffer:
I think this model is a better fit for the model zoo at [TuringExamples](https://github.com/TuringLang/TuringExamples). Could you slim this down into a `.jl` file with just the data preparation, the model, and the sampling step without any I/O stuff like plotting or printing? After, can you submit that as a PR at the [TuringExamples](https://github.com/TuringLang/TuringExamples) repo.

trappmartin:
Can we close this PR?

cpfiffer:
Yeah, I think so.

PR #46, opened 2019-05-30
[WIP] Add MMSB Tutorial

saumyagshah:


xukai92:
OK you won't be able to sample from a model with discrete random variables using pure HMC. Try to use PG or the Gibbs interface.

cpfiffer:
To kind of echo what @xukai92 said, the issue is that 

```julia
z_nm ~ Categorical(pi[n])
z_mn ~ Categorical(pi[m])
```

are discrete variables, so they need to be sampled using one of the discrete samplers. I'd try a sample call like the one from the [HMM](https://turing.ml/tutorials/4-bayeshmm/) tutorial:

```julia
g = Gibbs(1000, HMC(2, 0.001, 7, :m, :T), PG(20, 1, :s))
c = sample(BayesHmm(y, 3), g);
```

In your case, you'd pass `:z_nm` and `:z_mn` to `PG` and everything else to `NUTS`.

saumyagshah:
The chain did draw samples successfully but my reconstructed graph seems quite inaccurate. Does this require more iterations to run? This already ran for about an hour or so for 1000 iterations.

cpfiffer:
I think 1,000 iterations is probably not enough for 167 parameters. A visual inspection of the trace plots would suggest that few to none of these have converged. Plus, I think that the Dirichlet prior doesn't work super well with `HMC` ([link](https://arxiv.org/abs/1010.3436)). 

I wonder what happens if you sample with just `PG`? Probably won't get very far in terms of convergence, but it might be interesting to see.

saumyagshah:
I sampled used the PG sampler. It was fast but it gives hard assignments for the clusters (which makes sense). Also tried the HMC with 4000 iterations but it seems that we still have not reached convergence. Should I try using a smaller graph or try some other sampling algorithm like SGLD? Maybe I could sample the distribution differently as described in the paper that you linked though I am not sure how effective it would be.

cpfiffer:
Why not both? Try SGLD first and then reduce the graph size and go back to the original sampler specs with HMC.

saumyagshah:
SGLD would not be an appropriate sampler for this case since it is a constrained problem. I tried changing the model definition a bit according to [this](https://github.com/darthsuogles/mmsb) MMSB implementation. Even after reducing the graph size, I am still not able to obtain convergence. Can changing the parameters of the samplers or increasing the number of iterations make a difference?

cpfiffer:
First, I think you should first try a 2x2 to make sure the model works. 

Second, you're getting a ton of numerical errors with HMC. Try reducing the step size. You might also try running this on the master branch since there is a ton of stuff that's happened recently.

Third, you are getting this domain error which you should probably correct:

```julia
 DomainError with -2.220446049250313e-16:
log will only return a complex result if called with a complex argument. Try log(Complex(x)).  
```

I would try making the `Bernoulli` call in your model be lower-bounded at zero: 

```julia
val = max((pi[n])'*eta*pi[m], 0)
graph[n, m] ~ Bernoulli(val)
```

saumyagshah:
I think the model is working correctly since the graph with N=2, K=2, and N=3, K=2  is converging. I ran the latter 3 times and it perfectly predicted the generated data. Maybe, for larger graphs, we would need to use VI as MCMC is too slow at convergence. Keeping the number of iterations same, the prediction was decent for N=4, K=2. Do you think I should increase the size and run this for more iterations? I thought it would get too long for a tutorial.

cpfiffer:
It's interesting that it mostly converges after ~15k iterations. There's probably room to expand either N or K, if it gets interesting results.

saumyagshah:
Closing this and sending a PR to [TuringExamples](https://github.com/TuringLang/TuringExamples).

PR #47, opened 2019-09-10
 Tutorial for ADVI

torfjelde:
Since ADVI is quite different from everything else we have in Turing I figured I'd give it a proper introduction, but it might be a bit too much to put in a tutorial. Alternatively, I could skip most of the VI section and point the user to the ADVI paper? Love to hear what you guys think.

[nbviewer link](https://nbviewer.jupyter.org/github/torfjelde/TuringTutorials/blob/tor%2Fadvi-tutorial/8_VariationalInference.ipynb). You might have to add `?flush_cache=true` to the end to get the most recent version.


trappmartin:
Nice work. 

I think the current version is way too mathy. One solution would be to turn it into two separate tutorials. The first one more hands-on like the ADVI in Turing.jl section (+ Motivation and VI section) and a second one which explains the mathematical details. Something like a ADVI in detail tutorial for interested readers. Most people will probably be mainly interested in reading how to use the feature but it would be a pitty to throw the great explaination of VI away.

torfjelde:
> I think the current version is way too mathy. One solution would be to turn it into two separate tutorials. The first one more hands-on like the ADVI in Turing.jl section (+ Motivation and VI section) and a second one which explains the mathematical details. Something like a ADVI in detail tutorial for interested readers. Most people will probably be mainly interested in reading how to use the feature but it would be a pitty to throw the great explaination of VI away.

Thanks! And I think agree. Maybe we could move redundant parts into an appendix of sorts? Initially I thought that a "reading guide" would help a lot but re-ordering things would be probably be better. Could also change the reading guide to:

- "SHOW ME THE CODE!" → skip to [ADVI in Turing.jl](#ADVI-in-Turing.jl)
- "I don't know any of this, but I'd love to learn more!" → skip to [next section]()
- "I'm fairly familiar with VI, but ADVI? Sorry, what?" → skip to [Automatic Differentiation Variational Inference (ADVI)]()
- "VI and ADVI is so simple bro, like, whatever." → skip to [ADVI in Turing.jl]()
- "I could easily implement this myself. Why am I even looking at this?" → https://turing.ml/dev/docs/contributing/guide (he he he)

or something:)

cpfiffer:
I would put all (or most, anyway) of the math in a new document on the website, actually -- I don't really think it's tutorial material, but it might be interesting to a handful of people who are very interested in how VI works.

I think for this tutorial you're really underplaying the strength of VI by spending a lot of time on a pretty trivial example. The first section "ADVI in Turing.jl" should really just answer the question "what the hell am I supposed to type to get a number out of this thing".

The second tutorial should be a high-dimensional linear regression with a largeish sample size. Take the opportunity to show how long it would take using traditional MCMC vs. VI, so people are sold from the get-go. Don't include the MNIST tutorial -- it should be totally separate.

I'd recommend picking some random dataset from [RDatasets.jl](https://github.com/JuliaStats/RDatasets.jl) and picking a column (the one that seems important) and just regression everything on it. Try to find something with lots of rows and columns.

I started a goofy [side project](https://github.com/cpfiffer/BayesModels) I have that reduces linear regression models to a one-liner like 

```julia
chain = blm(@formula(y ~ x + x*z + m*q), df)
```

but in a lot of cases, typical models have either a high number of samples or high dimensionality (or both) and so MCMC is often a pretty poor performer. I might actually like to just see how well VI works on linear models, and maybe I can just build VI into my project as a default. You've already got a rapt audience.

torfjelde:
> I would put all (or most, anyway) of the math in a new document on the website, actually -- I don't really think it's tutorial material, but it might be interesting to a handful of people who are very interested in how VI works.

Fair enough :+1: Initially I didn't intend for the VI/ADVI part without code to be that long, but as I wrote I realized it needed to be to be of any use. Now, when I see how long it's become, I very much agree with you guys on the fact that it is best kept separately and merely referenced or something in this tutorial.

> I think for this tutorial you're really underplaying the strength of VI by spending a lot of time on a pretty trivial example.

I picked a very trivial example for two main reasons:
1. Closed form posterior exists which we can compare VI to. This demonstrates that it _can_ be a really good approximation.
2. No focus is put on the model and the method itself is what is presented.

> The first section "ADVI in Turing.jl" should really just answer the question "what the hell am I supposed to type to get a number out of this thing".

Hmm, can you elaborate a bit? You mean just start out with

```julia
# construct model
m = model(x);

# ADVI
advi = ADVI(1, 1000)
q = vi(m, advi);

# sample from variational posterior
samples = rand(q, num_samples)
```

and then follow up with an explanation of what the code does instead? In contrast to explain as I use.

> The second tutorial should be a high-dimensional linear regression with a largeish sample size. Take the opportunity to show how long it would take using traditional MCMC vs. VI, so people are sold from the get-go. Don't include the MNIST tutorial -- it should be totally separate.

Yeah that's what I was planning on doing with Linear Regression using ARD that I have as a TODO at then end. I was planning on running such a 250-dim example where I'd expect ADVI to blow "exact" methods out of the water.

cpfiffer:
> and then follow up with an explanation of what the code does instead? In contrast to explain as I use.

Basically, yeah. It's a lot easier to be intimidated or forgetful when really brief code like those three lines is spread out. You should expose them to basically what their code should look like when they write it. It also makes it easier to copy/paste.

Then when you are explaining the relevant pieces, you might call back to the line you are referencing a markdown cell with just that relevant line. Constant reinforcement helps people learn better.

> I picked a very trivial example for two main reasons:
> 
> 1. Closed form posterior exists which we can compare VI to. This demonstrates that it can be a really good approximation.
> 2. No focus is put on the model and the method itself is what is presented.

True enough -- I actually quite like the comparison to the closed form solution, but I think that it muddies the waters a bit, because you're trying to show two things at once (how to use it and how VI stacks up to the analytic solution). Maybe a quick fix is to just add a section right before the conjugacy model that just shows how to do VI with the `gdemo` model, so people just get the hang of the technique. In there you can explore what each line means (`ADVI`, `vi`, etc.).

Afterwards you can show how VI compares to the closed-form solution, which I think is important and useful but it should be separate from the basic mechanics.

Tutorials like this one are meant basically to get people limping along with the code, in my opinion. I think really good ones tend to answer two questions:

1. What do I type to make the thing work?
2. What is something really cool or useful I can do with whatever this thing is?

With the preliminary draft, I think you've mostly answered part one but it's not as clear as it could be, and then the linear model you've got planned should answer part two. The closed-form comparison stuff is really important in that it adds more detail to this method, so I'd keep it stuck in between the "show people what to type" section and the "here's something totally rad" section.

I normally wouldn't care that much if this were any of the other Turing tutorials, but since this is for a new method an it's a less common method of inference, I think we should be a little more delicate with users than normal. 

I like your writing style and I think you have a very good handle on conveying complex ideas. I think so far this is a pretty excellent document and I'm pleased to have something we can show to people.

torfjelde:
> Basically, yeah. It's a lot easier to be intimidated or forgetful when really brief code like those three lines is spread out. You should expose them to basically what their code should look like when they write it. It also makes it easier to copy/paste.

Sounds like a good idea; I'll try to do that :+1:

> With the preliminary draft, I think you've mostly answered part one but it's not as clear as it could be, and then the linear model you've got planned should answer part two. The closed-form comparison stuff is really important in that it adds more detail to this method, so I'd keep it stuck in between the "show people what to type" section and the "here's something totally rad" section.

I'll restructure and rewrite to similiar to your suggestion :+1: Thank you for the feedback; it's really helpful!

> I like your writing style and I think you have a very good handle on conveying complex ideas. I think so far this is a pretty excellent document and I'm pleased to have something we can show to people.

Thanks man:)

torfjelde:
AIGHT. This should now be ready for another look-over.

The maths-section is still there, but the code-section is now self-contained, so it's just a matter of resolving whether or not to drop the mathsy section or alternatively put in the appendix. I think the general consesus was to remove it which is fine by me! Everyone who wants to remove say "aye" (or give a :+1: ).

trappmartin:
I’ll have a look at it. But I’m in favour of not dropping the math but putting it into an appendix or so. I think it’s extremely useful for users that want to know about the details.

yebai:
> I’ll have a look at it. But I’m in favour of not dropping the math but putting it into an appendix or so. I think it’s extremely useful for users that want to know about the details.

Maybe move the detailed VI derivations to a separate docs page (e.g. under the "For Developers" section on the docs website) and refer that it.

cpfiffer:
I like the updates to the tutorial! I've only got two suggestions:

1. I'd second @yebai's suggestion that the math bits be moved to the docs. 
2. For the mean +/- standard deviation plots you do, can you add a legend or something so people know what the range/blue dots are at a quick glance?

I'm fine with merging after the two above points are addressed.

torfjelde:
Have now moved the mathy section over to https://github.com/TuringLang/Turing.jl/pull/1178

torfjelde:
Anything left for this to get merged?

yebai:
Please create a PR on Turing and add VI to the tutorial list.

PR #48, opened 2019-09-11
Update DataFrames code & general clean up

nilshg:
I couldn't work out how to edit the existing notebook on GitHub so had to deleted and then upload the new version unfortunately, which makes the diff very large unfortunately...

Basically this PR just tidies the code a bit, mainly in terms of some DataFrames operations which are now deprecated (`deletecols!` and some of the indexing). Also a few typos fixed.


nilshg:
Have changed the wording on the Normal priors, I've tried MLDataUtils but couldn't get `rescale!` to give sensible results unfortunately...

trappmartin:
Hm, could you try `rescale` again. It does almost exact what the code does atm. You should be able to do the following:
```julia
train = (train .- mean(train, dims=1)) ./ std(train, dims=1)
test = (test .- mean(test, dims=1)) ./ std(test, dims=1);
```
changes to
```julia
m, v = rescale!(train, obsdim = 1)
test .-= m'
test ./= v'
```
which should be anyhow the way it should be. Normalising each matrix independently is in my opinion anyhow a bug.

yebai:
Closed in favour of https://github.com/TuringLang/TuringTutorials/pull/280

PR #55, opened 2020-03-26
Add multinomial logistic regression example

DilumAluthge:
This pull request adds a multinomial logistic regression to the list of tutorials.

It closely mirrors the logistic regression tutorial.


REQUIRE
```diff
@@ -11,3 +11,4 @@ StatsPlots
 Flux#master
 StatsFuns
 GLM
+NNlib
```
devmotion: Could we just replace the REQUIRE file with a Project.toml file? REQUIRE files were replaced with Project.toml files quite a while ago.

DilumAluthge: Done!

cpfiffer:
Short and sweet. I like it! Thanks for the PR! It'll take a bit for this to show up on the site, since there's some bugs in other tutorials that need to be fixed (https://github.com/TuringLang/Turing.jl/pull/1180).

PR #71, opened 2020-05-30
end of concrete_solve

ChrisRackauckas:
After seeing how these tutorials turned out, I took it upon myself to figure out how to make it so that DiffEq code could use adjoints with literally no change and is safe. Turns out it works. This PR removes `concrete_solve` and makes it so that it's all `solve` usage, since `concrete_solve` is now deprecated in DiffEqBase v6.36. The manifest might need an update, and my computer is the absolute worst case right now for that so I just directly edited the rendered results.

@vaibhavdixit02


cpfiffer:
Thanks! I imagine using `solve` should change the results too, do you want me to rerun this?

Vaibhavdixit02:
@cpfiffer we would have the same problem and it might need to be run multiple times for it to look good. Also, something's broken in OrdinaryDiffEq that prevents the run of the `zygote` backend problems. I will put in a fix in some time

ChrisRackauckas:
>Also, something's broken in OrdinaryDiffEq that prevents the run of the zygote backend problems.

What do you mean?

cpfiffer:
Can we just set a seed? It's incredibly expensive and time consuming to run this thing and I'm not interested in running and rerunning it every time we update something to see if it's mixing well. There are too many other tutorials to keep track of and they should all mostly work whenever they're run.

ChrisRackauckas:
FWIW, the Turing.jl v0.13 update only broke the first example which is pretty quick, so it would be good to isolate that out and find out what happened in the latest release.

cpfiffer:
I mean more that I should be able to rerun this notebook locally and it should mix the same as the markdown file, and if that's not the case, something needs to be modified. I thought @Vaibhavdixit02's comment was that if I run this twice, one might be fine and the other might be garbage.

ChrisRackauckas:
The chains sometimes do wonky stuff, but that's something that should get fixed instead of glossed over. One thing that could help is starting at the MAP estimate.

BTW @Vaibhavdixit02 , why did you change the default parameters of the LV equation? That might've made it a bit less stable of an example.

Vaibhavdixit02:
>why did you change the default parameters of the LV equation?

I didn't, I thought you wanted to run it with these values since they are off from the mean of the distributions purposefully.

ChrisRackauckas:
`p = [2.2, 1.0, 2.0, 0.4]` aren't the parameters I started on. It's a somewhat odd parameter set because it doesn't capture the cyclic nature of the solution very well.

Vaibhavdixit02:
@cpfiffer it should be good to go now

cpfiffer:
Thanks, folks!

ChrisRackauckas:
@Vaibhavdixit02 I think you pushed the wrong version: https://github.com/TuringLang/TuringTutorials/blob/master/10_BayesianDiffEq.ipynb

(Having this as notebooks is a major PITA)

cpfiffer:
Yeah, just noticed that. I too hate having these as notebooks.

Vaibhavdixit02:
>I think you pushed the wrong version:

Do you mean because of the incomplete runs?

cpfiffer:
There's a bunch of undefined errors in the notebook.

ChrisRackauckas:
There's a bunch of errors from using an old OrdinaryDiffEq version, the first cell changes the parameters but didn't update the plot and the updated parameters are only used in the first example, so the other examples need to be re-ran. Etc.

Vaibhavdixit02:
@cpfiffer that should work with the new release of OrdinaryDiffEq

Vaibhavdixit02:
> updated parameters are only used in the first example, so the other examples need to be re-ran. Etc.

They will be reran for deployment is what I understood?

cpfiffer:
We haven't set up a GitHub action to do it, so I typically rerun them on my machine and then push the updated markdown files to the repo. Super low-tech.

Vaibhavdixit02:
You should be good to go for that @cpfiffer, I saw that it runs quite a lot faster on your end so didn't run the entire notebook again purposefully

cpfiffer:
Cool, works for me. I'm running it now and I'll post here if there's issues.

ChrisRackauckas:
Alright, yes, it should work when it re-runs, so 🤞

PR #72, opened 2020-05-31
Revert "end of concrete_solve"

cpfiffer:
Reverts TuringLang/TuringTutorials#71


cpfiffer:
Decided it's not worth it to revert, I think this will be fixed in a separate PR.

Issue #73, opened 2020-06-02
Fix unstandardize function.


GarrettMooney:
The `mean` and `std` functions need to be switched in the `unstandardize` function.

This function is found in both the [linear regression](https://github.com/TuringLang/TuringTutorials/blob/master/markdown/5_LinearRegression.md) and the [VI](https://github.com/TuringLang/TuringTutorials/blob/master/markdown/9_VariationalInference.md) tutorials.

I'd be glad to submit a PR but I'll need a bit more guidance as to best practices for this repo.

cpfiffer:
I'm not sure what you mean switched. Do you mean adding the standard deviation? Or do you mean switching the order of operations? If it's the first, I don't think that's right -- you add the mean, then multiply by the standardization. If it's the second, it doesn't matter -- multiplication is commutative.

devmotion:
IMO @GarrettMooney is right, the inverse function to the standardization in `standardize` should be
```julia
function unstandardize(x, orig)
    return x .* std(orig, dims=1) .+ mean(orig, dims=1)
end
```
instead of
```julia
function unstandardize(x, orig)
    return (x .+ mean(orig, dims=1)) .* std(orig, dims=1)
end
```
They are not equivalent since in the current IMO incorrect case the original mean is also scaled with the standard deviation (whereas it should only be a translation afterwards).

GarrettMooney:
I'm realizing that I did not communicate the issue very well but I think David summarized it nicely. 

For an additional intuition check, the current version of the [linear regression notebook](https://github.com/TuringLang/TuringTutorials/blob/master/5_LinearRegression.ipynb) shows that MPG was originally near 20 but the `unstandardize` function transforms predictions for MPG to be closer to 120.

cpfiffer:
Ah, I see. We'd be happy to have a PR to fix this.

PR #75, opened 2020-06-08
Fix linear regression tutorial

devmotion:
This PR fixes https://github.com/TuringLang/TuringTutorials/issues/73. Instead now MLDataUtils is used for splitting and rescaling the data sets, as in the logistic regression tutorial.

Additionally, the data leakage is fixed (standardization was computed based on the whole data set and not only the training data) and the Turing model is corrected so that it corresponds to the explanation in the text (the standard deviation of `Normal` and `MvNormal` was specified in the model definition which should have been the variance according to the text). Furthermore, I removed all unneeded model inputs (`n_obs`, `n_vars`, and `::Type{T}` are not needed). Additionally, the prediction of the Bayesian model is now computed as the mean observation for all Markov samples instead of the observation for the model with the mean of the Markov samples as parameters.

With the chosen seed the Bayesian model and OLS perform about the same on training and test data set (the Bayesian model is slightly better on the test data set but slightly worse on the training data set - I doubt though that this is consistent for different seeds).

I couldn't reproduce https://github.com/TuringLang/Turing.jl/issues/1313 with the updated tutorial, so I assume https://github.com/TuringLang/Turing.jl/issues/1313 is fixed by this PR.


devmotion:
<p><code>testset[:, target]</code>  copies <code>testset.MPG</code>  whereas <code>testset[!, target]</code>  doesn't (it just returns <code>testset.MPG</code>  in our case), see http://juliadata.github.io/DataFrames.jl/stable/lib/indexing/. I wanted to avoid the allocations here.</p>

---

 <a href='https://app.reviewnb.com/TuringLang/TuringTutorials/pull/75/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

PR #76, opened 2020-06-10
Add measurement noise to ode data and restructure sde example

Vaibhavdixit02:


devmotion:
The notebook is invalid, isn't it? It seems it contains a lot of merge conflicts, and can't be displayed by ReviewNB, Github, and nbviewer.

Vaibhavdixit02:
@devmotion yes it looks like it. I don't really know how that happened, must have been some weird interaction of git and the notebook because it was working for me when I had pushed. I'll fix this in a bit sorry for the trouble. 😓

PR #82, opened 2020-08-09
Updated version

arnaudmgh:
In collaboration with @Vaibhavdixit02.

* Corrected an inversion in parameters gamma and delta that produced wrong results
* Added plots of retrodicted time courses from the posterior
* Added 3 chains per problem, one using a single thread with `mapreduce`, one using Threads
* As the SDE example as a few problems, numerical and conceptual, we replaced with a DDE problem that run fast and worked. 
* Moved DiffEqBayes section lower, as now all functions of DifferentialEquations.jl and Turing.jl are composable.
* The problem using sensitivity runs the slowest - probably because it is made for larger problem and has a bigger overhead... 
* Added a small comment specifying that ~ is the tilde character and not \sim.


Vaibhavdixit02:
@ChrisRackauckas any comments?

ChrisRackauckas:
Where did the SDEs go?

Vaibhavdixit02:
>Where did the SDEs go?

I had asked to remove the example for now since it was brought up on several occasions that the likelihood defined based on the mean was not very informative. @arnaudmgh I don't recall if you tried the SDE example with `NUTS(0.45)`?

ChrisRackauckas:
>I had asked to remove the example for now since it was brought up on several occasions that the likelihood defined based on the mean was not very informative. @arnaudmgh I don't recall if you tried the SDE example with NUTS(0.45)?

Then just change the likelihood?

ChrisRackauckas:
For SDEs, a quasi-likelihood approach seems to do well:

```julia
using DifferentialEquations, Plots, Optim, DiffEqSensitivity, Turing, MCMCChains, StatsPlots
function lotka_volterra!(du,u,p,t)
  x,y = u
  α,β,γ,δ = p
  du[1] = dx = α*x - β*x*y
  du[2] = dy = δ*x*y - γ*y
end
u0 = [1.0,1.0]
tspan = (0.0,10.0)
function multiplicative_noise!(du,u,p,t)
  x,y = u
  du[1] = p[5]*x
  du[2] = p[6]*y
end
p = [1.5,1.0,3.0,1.0,0.1,0.1]
prob = SDEProblem(lotka_volterra!,multiplicative_noise!,u0,tspan,p)
ensembleprob = EnsembleProblem(prob)
@time data = solve(ensembleprob,SOSRI(),saveat=0.1,trajectories=1000)
plot(EnsembleSummary(data))

Turing.setadbackend(:forwarddiff)
@model function fitlv(data)
    σ ~ InverseGamma(2,3)
    α ~ truncated(Normal(1.3,0.5),0.5,2.5)
    β ~ truncated(Normal(1.2,0.25),0.5,2)
    γ ~ truncated(Normal(3.2,0.25),2.2,4.0)
    δ ~ truncated(Normal(1.2,0.25),0.5,2.0)
    ϕ1 ~ truncated(Normal(0.12,0.3),0.05,0.25)
    ϕ2 ~ truncated(Normal(0.12,0.3),0.05,0.25)
    p = [α,β,γ,δ,ϕ1,ϕ2]
    prob = SDEProblem(lotka_volterra!,multiplicative_noise!,u0,(0.0,10.0),p)
    predicted = solve(prob,SOSRI(),saveat=0.1)

    if predicted.retcode != :Success
        Turing.acclogp!(_varinfo, -Inf)
    end
    for j in 1:length(data)
        for i = 1:length(predicted)
            data[j][i] ~ MvNormal(predicted[i],σ)
        end
    end
end;
model = fitlv(data)
chain = sample(model, NUTS(0.25), 5000, init_theta = [1.5,1.3,1.2,2.7,1.2,0.12,0.12])
plot(chain)
savefig("plot.png")
```

plot

The idea behind this is that likelihood function is noisy, but because we're also doing a form of stochastic optimization it's okay to not have to use ensembles because the ensemble sampling is in the MCMC process itself. HMC NUTS can in theory be slightly biased, but the SGHMC implementation doesn't seem to be working right now (and that has a bias correction for this case), so this would have to do for now. the NUTS sampler is a bit sensitive to the stochastic optimization since the gradient is then changing with every calculation, so to handle that the target acceptance rate is lowered and that stabilizes it a bit. You can still see that in some parameter directions this leads to small steps (so does anyone know how to manually force that to be larger?), but in general it works quite well. For a serious run I would've done a longer burn in because you can see the noise parameters are still marching to the true solution, but as this is it's not too computationally heavy, so the tutorial could just say, boost a bunch of the stats if you want the full fit.

Let me know if you need any help writing this version.

ChrisRackauckas:
And just for future reference, a version which doesn't use the quasi-likelihood approach:

```julia
using DifferentialEquations, Plots, Optim, DiffEqSensitivity, Turing, MCMCChains, StatsPlots, Statistics
function lotka_volterra!(du,u,p,t)
  x,y = u
  α,β,γ,δ = p
  du[1] = dx = α*x - β*x*y
  du[2] = dy = δ*x*y - γ*y
end
u0 = [1.0,1.0]
tspan = (0.0,10.0)
function multiplicative_noise!(du,u,p,t)
  x,y = u
  du[1] = p[5]*x
  du[2] = p[6]*y
end
p = [1.5,1.0,3.0,1.0,0.15,0.15]
prob = SDEProblem(lotka_volterra!,multiplicative_noise!,u0,tspan,p)
sol = solve(prob)
plot(sol)
ensembleprob = EnsembleProblem(prob)
@time sol = solve(ensembleprob,SOSRI(),saveat=0.1,trajectories=1_000)
truemean = mean(sol,dims=3)[:,:]
truevar = var(sol,dims=3)[:,:]
Turing.setadbackend(:forwarddiff)
@model function fitlv(data)
    σ1 ~ InverseGamma(2,3)
    σ2 ~ InverseGamma(2,3)
    α ~ truncated(Normal(1.5,0.3),0.5,2.5)
    β ~ truncated(Normal(1.2,0.3),0.5,2)
    δ ~ truncated(Normal(3.0,0.3),2.2,3.5)
    γ ~ truncated(Normal(1.0,0.3),0.5,1.5)
    ϕ1 ~ truncated(Normal(0.12,0.3),0.1,0.3)
    ϕ2 ~ truncated(Normal(0.12,0.1),0.1,0.3)
    p = [α,β,δ,γ,ϕ1,ϕ2]
    prob = SDEProblem(lotka_volterra!,multiplicative_noise!,u0,(0.0,10.0),p)
    ensemble_predicted = solve(EnsembleProblem(prob),SOSRI(),saveat=0.1,trajectories=20)
    predicted_means = mean(ensemble_predicted,dims=3)[:,:]
    predicted_vars  = var(ensemble_predicted,dims=3)[:,:]
    for i = 1:size(predicted_means,2)
        data[1][:,i] ~ MvNormal(predicted_means[:,i], σ1)
        data[2][:,i] ~ MvNormal(predicted_vars[:,i], σ2)
    end
end;
model = fitlv([truemean,truevar])
chain = sample(model, NUTS(0.35), 500; progress=true)
plot(chain)
savefig("plot.png")
```

but it's a bit more expensive in both compute and data so the other is preferred.

ChrisRackauckas:
I don't have all that much time right now, so passing the "good enough" is, well, good enough. I agree the priors can be adjusted and all of that though, but will have to leave that to the future.

Vaibhavdixit02:
I think this should be good to go in now

devmotion:
Can we maybe change the logging level for this notebook to avoid all the warnings from AdvancedHMC while tuning the step size?

cpfiffer:
Yeah, these should ultimately be run by CI, but I haven't gotten around to setting it up yet. For the moment, I will rerun the markdown file on my machine.

I think the only remaining things to address here are @devmotion's comments above.

Vaibhavdixit02:
Thanks for addressing the review @arnaudmgh. cc: @cpfiffer

Vaibhavdixit02:
Bump @arnaudmgh ?

PR #83, opened 2020-08-23
Remove old markdown for DiffEq tutorial

Vaibhavdixit02:
The rendered version https://turing.ml/dev/tutorials/10-bayesiandiffeq/ seems to be the old markdown still present


devmotion:
@cpfiffer Apparently the tutorial was renamed and the current setup does not remove the old markdown files automatically. Hence on the webpage the old tutorial is still present: https://turing.ml/dev/tutorials/10-bayesiandiffeq/ and the updated one is available here: https://turing.ml/dev/tutorials/10-diffeq/

Now, one could remove the old markdown files (I know that this won't affect the current and previous versions of the webpage but I assume it should fix the problem for future builds) but I think it might be better to just rename the tutorial to 10-BayesianDiffEq so that old/current links are still valid. What's your opinion?

cpfiffer:
I'd just rename it.

PR #84, opened 2020-09-01
Use Flux.destructure to create network from parameters

theogf:
Use the method Flux.destructure to create a vector of parameters and reconstruction method from an existing network.
It is much cleaner and removes the need for the extra section for general networks.


devmotion:
That's definitely much nicer than before, the only disadvantage of this approach (which, however, already existed in the previous implementation) is that it does not allow to access the weights of different layers in an easy way and hence it is not really possible to specify different priors for different parts of the network. I would assume (haven't tried though) that [ComponentArrays](https://jonniedie.github.io/ComponentArrays.jl/dev/examples/DiffEqFlux/) might be useful if one is interested in such a case.

theogf:
For different priors can't the same method be used in the same way by dividing each network separately?
```
nn1 = Chain(Dense(3,2), Dense(2, 4))
nn2 = Chain(Dense(4,1))
nn = Chain(nn1, nn2)
w1, re1 = Flux.destructure(nn1)
w2, re2 = Flux.destructure(nn2)
function nn_forward(x, theta1, theta2)
  nn = Chain(re1(theta1), re2(theta2)
  nn(x)
end
```

devmotion:
Only if you can divide it up in the way that you want to split up your priors. E.g., it wouldn't allow you to specify different priors for the weights and offset terms, i.e., `W` and `b` in each `Dense` layer.

theogf:
Fair enough! But for a Turing tutorial is it really necessary to go that far into details? Maybe I can add a note that this **one** way to do it, but that for a more fine grained model other approaches are possible.

devmotion:
I'm very much in favour of this PR. I just hink it would be good to keep in mind that it might not always work. Maybe it would be good to add such a comment, otherwise people might think it's the only way.

cpfiffer:
I'm happy to merge too if you want to add a comment or something. I honestly have no preference on that -- I'm happy with this as-is, but David is generally more careful than me.

torfjelde:
>  ComponentArrays might be useful if one is interested in such a case.

Btw, I think this is a really nice and general approach. First saw it here https://julialang.zulipchat.com/#narrow/stream/237519-turing/topic/ComponentArrays.20.2B.20Turing.20.2B.20Flux/near/202590251 (thanks @jonniedie !)

In the VI-tutorial, this means that

```julia
function getq(θ)
    offset = 0
    L = LowerTriangular(reshape(@inbounds(θ[offset + 1: offset + d^2]), (d, d)))
    offset += d^2
    b = @inbounds θ[offset + 1: offset + d]
    
    # For this to represent a covariance matrix we need to ensure that the diagonal is positive.
    # We can enforce this by zeroing out the diagonal and then adding back the diagonal exponentiated.
    D = Diagonal(diag(L))
    A = L - D + exp(D) # exp for Diagonal is the same as exponentiating only the diagonal entries
    
    b = to_constrained ∘ Shift(b; dim = Val(1)) ∘ Scale(A; dim = Val(1))
    
    return transformed(base_dist, b)
end
```
converts into
```julia
# Example array that we'll use as a template for destructuring/restructuring.
proto_arr = ComponentArray(
    L = zeros(d, d),
    b = zeros(d)
)
proto_axes = proto_arr |> getaxes
num_params = length(proto_arr)

function getq(θ)
    L, b = begin
        # Construct array from vector `θ` and "template" axes `proto_axes`,
        # then unpack.
        @unpack L, b = ComponentArray(θ, proto_axes)
        LowerTriangular(L), b
    end
    # For this to represent a covariance matrix we need to ensure that the diagonal is positive.
    # We can enforce this by zeroing out the diagonal and then adding back the diagonal exponentiated.
    D = Diagonal(diag(L))
    A = L - D + exp(D) # exp for Diagonal is the same as exponentiating only the diagonal entries
    
    b = to_constrained ∘ Shift(b; dim = Val(1)) ∘ Scale(A; dim = Val(1))
    
    return transformed(base_dist, b)
end
```
It's slightly longer, but much more readable + this is a "trivial" example. Once you have more parameters, etc. the first impl gets awful (e.g. the example in the chat I linked above). And if we're going to use this approach, then I'd prefer to just use it in all tutorials rather than one or two depending on the cases.

In particular, I think people not familiar with Julia will be happy to see that there's an easy way and general way of doing this:)

theogf:
Shouldn't we propose both approaches? I agree the ComponentArrays approach is very nice for controlling everything. But it's also nice, to get a premade NN, and be able to just transform it into a BNN with a few lines :D

trappmartin:
What is the status? Can this be merged?

theogf:
It would be nice to incorporate all the different comments (maybe at the end of the tutorial) on the alternative ways to do this! I don't have time right now and if someone wants to chime in to do it that's fine with me :)

yebai:
@xukai92 Would you like to take a look, and update the `jmd` version of `BayesianNN`?

xukai92:
I had a version of BNN with the `destructure` trick in my JuliaCN 2020 talk (https://github.com/xukai92/juliacn-2020/blob/master/bnn.jl#L208-L250)---I will just go and update the `jmd` version using that trick.

PR #85, opened 2020-09-17
Fix issues in tutorial of multinomial logistic regression

devmotion:
This PR fixes https://github.com/TuringLang/Turing.jl/issues/1409 and some additional issues such as leaking information from the test data set (the rescaling was performed using mean and standard deviation of the whole data set) and duplicate corner plots (the second one was just a duplicate of the first one). Moreover, the model and the prediction is simplified.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -26,9 +26,9 @@ version = "0.3.3"
 
 [[Adapt]]
 deps = ["LinearAlgebra"]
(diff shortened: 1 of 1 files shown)


cpfiffer:
Thanks for the fix, looks good!

devmotion:
If you push the rebuilt markdown files, you can manually retrigger a build of the docs in Turing to update the webpage.

cpfiffer:
Will do.

Issue #86, opened 2020-09-23
Tutorials are largely outdated


torfjelde:
A lot of the tutorials are unfortunately outdated due to recent improvements in MCMCChains, e.g. https://github.com/TuringLang/Turing.jl/issues/1411.

I'm opening an issue here so as to bring attention to the, well, _issue_.

devmotion:
I think maybe it would help to 1) add separate environments for different tutorials (so people can reproduce it even though it is based on an older version of some package) and 2) use CI tests and builds for the tutorials.

torfjelde:
100%. But for (2), we likely have to move away from Jupyter notebooks, right? Which, honestly, would be it's own benefit.

torfjelde:
And regarding (1), I would say that _ideally_ we'll always have the tutorials work with the most recent release of Turing, since people will look to this for inspiration when writing their own code.

Maybe we should run `Pkg.status()` at the beginning or end of every notebook too, so that people can view the dependencies.

devmotion:
> Maybe we should run `Pkg.status()` at the beginning or end of every notebook too, so that people can view the dependencies.

That's how it is done in SciML: https://github.com/SciML/SciMLTutorials.jl An appendix is added automatically to every tutorial which shows the packages and Julia version that was used to run it

torfjelde:
> That's how it is done in SciML

Nice :+1:  And they use `jmd` as the "base" file?

devmotion:
Yes. And then notebook, HTML, and PDF files are generated with Weave.

torfjelde:
Super-dope. Would it be a good idea to just try and duplicate their approach?

devmotion:
Yes, I think that would be great. In particular also thinking about https://github.com/SciML/RebuildAction which we use over at SciML to rebuild tutorials automatically (together with CompatHelper for individual environments this helps to figure out any possible issues).

torfjelde:
Yeah I just noticed that; awesome stuff! Cool, well I'm very down for doing that.

cpfiffer:
All of this is a great idea. I hate the way the tutorials currently work -- they're difficult to write in, difficult to automate, and often out-of-date. Switching to a more general `jmd` approach is something I'm all for.

trappmartin:
Super cool idea.

torfjelde:
I've now updated all the tutorials to work with the new system (which is in a separate branch `jmd-staging`, which we'll merge into master when everything is ready for a full update).

[...]

PR #87, opened 2020-09-24
[WIP] Implementation of #86

torfjelde:
Ref #86 and its discussion. This PR is WIP for converting TuringTutorials.jl into a similar structure as https://github.com/SciML/SciMLTutorials.jl. 

# TODOs
1. [ ] Convert tutorials to JMD format and make sure they're working.
2. [ ] Add proper build-script/CI.
   - Worth noting that we don't *need* this to merge the PR, as we can just manually generate the markdown files and push.


devmotion:
> Worth noting that we don't need this to merge the PR, as we can just manually generate the markdown files and push.

I'd suggest to keep this PR as simple as possible, and just focus on the Weave commands + converting all notebooks to jmd here (I guess one might be able to do this in a semi-automated fashion using Weave?). Then we should just run them locally, using the Weave scripts, before merging the PR (SciMLTutorials used a manual build process for years before switching to CI some months ago).

devmotion:
I would maybe go even further and not even touch the existing environment nor update the tutorials in this PR, since I guess there is a lot to update and it might become confusing quite quickly.

torfjelde:
> I would maybe go even further and not even touch the existing environment nor update the tutorials in this PR, since I guess there is a lot to update and it might become confusing quite quickly.

100% agree with this! (Though I did make a tiny change to the VI tutorial, but not planning on making any more!)
What I've done now is literally just go through the tutorial, added `results = "hidden"` where it's intended to now show output (`Weave.jl` doesn't seem to respect the `;` at the end of the block) and made sure that the tutorial runs with updated packages.

BUT I also realize that because I'm moving files around, etc. the review process is going to be horrible. 

Maybe it would be a good idea (maybe this is what you're also getting at) to do it in several PRs:
1. Only change the files to JMD format (without changing the file extension), e.g. adding `results = "hidden"`. 
2. Move all files to the `tutorials/...` folder and add `jmd` extension + remove all notebooks and clean the `markdown` folder, i.e. only move + delete files, but no actual changing of the files.
3. Add `Project.toml` to each of the tutorials and make sure tutorials are running, fixing accordingly.

Also, all of the above should probably be merged into a `staging` branch or something, and then we do a complete final merge to `master`.

Thoughts?

devmotion:
> Maybe it would be a good idea (this is what you're also getting at, right?) to do it in several PRs:

Yes, exactly, that's mainly what I had in mind.

> Also, all of the above should probably be merged into a staging branch or something, and then we do a complete final merge to master.

That's a good idea! Let's just make sure that every change to master before the branch is merged, goes into the staging branch as well.

So, yes, I think this sounds like a good plan :+1:

torfjelde:
Great! 
Btw, I don't need reviews for just moving files around, right? So should I just create such a branch with the files moved and deleted, and then simply make PRs to that.

torfjelde:
How's this @devmotion ? `jmd-staging` is a branch with all the files moved and deleted, but without any changes made to the contents + added `src/TuringTutorials` and cleaned up `Project.toml`.

tutorials/variational-inference/01_variational-inference.jmd
```diff
@@ -32,44 +31,32 @@ Random.seed!(42);
...
-
 ```julia
 @model model(x) = begin
     s ~ InverseGamma(2, 3)
     m ~ Normal(0.0, sqrt(s))
     for i = 1:length(x)
         x[i] ~ Normal(m, sqrt(s))
     end
-end
+end;
```
devmotion, line 55: Maybe better to use `results = "hidden"` here as well? Or just leave it as it is for now?

tutorials/variational-inference/01_variational-inference.jmd
```diff
@@ -127,58 +89,22 @@ By default, i.e. when calling `vi(m, advi)`, Turing use a *mean-field* approxima
...
-    ```
-    
-    Create an [`ADVI`](@ref) with the currently enabled automatic differentiation backend `ADBackend()`.
-
-
 To perform VI on the model `m` using 10 samples for gradient estimation and taking 1000 gradient steps is then as simple as:
 
 
-```julia
+```julia; results = "hidden"
```
devmotion, line 102: Shouldn't/Couldn't the semicolon at the end dropped in this case?

tutorials/variational-inference/01_variational-inference.jmd
```diff
@@ -190,124 +116,66 @@ And one significant advantage of using `vi` is that we can sample from the resul
...
-
-
 That's pretty close! But we're Bayesian so we're not interested in *just* matching the mean.
 Let's instead look the actual density `q`. 
 
 For that we need samples:
 
 
-```julia
+```julia; results = "hidden"
```
devmotion, line 150: Same here?

tutorials/variational-inference/01_variational-inference.jmd
```diff
@@ -190,124 +116,66 @@ And one significant advantage of using `vi` is that we can sample from the resul
...
-    Plots.PyPlotBackend()
-
-
-
-
 ```julia
 p1 = histogram(samples[1, :], bins=100, normed=true, alpha=0.2, color = :blue, label = "")
 density!(samples[1, :], label = "s (ADVI)", color = :blue, linewidth = 2)
-density!(collect(skipmissing(samples_nuts[:s].value)), label = "s (NUTS)", color = :green, linewidth = 2)
+density!(collect(skipmissing(samples_nuts[:s].data)), label = "s (NUTS)", color = :green, linewidth = 2)
```
devmotion, line 163: Is this change needed? IMO probably we should use something different here since we have all the array-conversion methods.

devmotion: Why can't we just state the parameter BTW?

tutorials/variational-inference/01_variational-inference.jmd
```diff
@@ -351,7 +219,7 @@ p_μ_pdf = z -> exp(logpdf(p_μ, (z - μₙ) * exp(- 0.5 * log(βₙ) + 0.5 * lo
 p1 = plot();
 histogram!(samples[1, :], bins=100, normed=true, alpha=0.2, color = :blue, label = "")
 density!(samples[1, :], label = "s (ADVI)", color = :blue)
-density!(collect(skipmissing(samples_nuts[:s].value)), label = "s (NUTS)", color = :green)
+density!(vec(samples_nuts[:s].data), label = "s (NUTS)", color = :green)
```
devmotion, line 222: Same question here?

devmotion (commented):
Looks OK, I added some minor comments.

In general, I think the `results = "hidden"` statements are applied a bit inconsistently here - shouldn't we remove all the semicolons that are used to suppress output currently if we add them? Or just not add them for now?

tutorials/variational-inference/01_variational-inference.jmd
torfjelde: So it actually seems like this is not intended? https://github.com/JunoLab/Weave.jl/issues/386

If so, we can just remove all the `results = "hidden"` as this was the reason why I added them.

tutorials/variational-inference/01_variational-inference.jmd
torfjelde: You mean just do `vec(samples_nuts[:s])`? If so, yeah! I did this a bit quickly and just assumed `vec` didn't work, but it's an `AxisArray` now so ofc.

tutorials/variational-inference/01_variational-inference.jmd
devmotion: This issue seems to be about a different problem though, doesn't it? It doesn't mention `results = "hidden"`, which according to the documentation is intended to hide the output?

tutorials/variational-inference/01_variational-inference.jmd
devmotion: Hmm is `samples_nuts` not a Chains object? I though either we should use something like `Array(samples_nuts[:s])` (or just `samples_nuts[:s]`) (maybe with `vec` if needed) or just `density!(samples_nuts, :s, ...)` (or something similar) since we define all these recipes in MCMCChains.

torfjelde: > density!(samples_nuts, :s, ...)

Wait, WHAT?! How come I don't know about this?! That's awesome :+1: I'll use those then !

devmotion: Haha make sure to check if/how it works first and don't rely on my vague memories :smile:

tutorials/variational-inference/01_variational-inference.jmd
torfjelde: I added `results = "hidden"` because `;` wasn't respected. If `;` is respected, then we none of the `results = "hidden"` I added are needed.

torfjelde: Made a PR with a fix, so I'll remove these `results = "hidden"` and wait for new relase.

devmotion: Great!

I'm still surprised that `results = "hidden"` doesn't actually suppress the output even though it is mentioned in the docs in this way.

devmotion: I generally think the semicolons are a bit ugly but let's stick with them if that's the only way to make it work reliably :shrug:

torfjelde: > I'm still surprised that results = "hidden" doesn't actually suppress the output even though it is mentioned in the docs in this way.

Wat. No it _does_ What did I say? :confused: 

> I generally think the semicolons are a bit ugly but let's stick with them if that's the only way to make it work reliably 

Yeaah, let's for now :+1:

devmotion (approved):

tutorials/variational-inference/01_variational-inference.jmd
torfjelde: It worked:)

PR #88, opened 2020-09-25
Bobbing and weaving in the new setup

torfjelde:
This PR attempts to set up a basic way for to generate tutorials to be render on the website.

### DIFF
--- a/src/TuringTutorials.jl
+++ b/src/TuringTutorials.jl
@@ -7,117 +7,120 @@ cssfile = joinpath(@__DIR__, "..", "templates", "skeleton_css.css")
 latexfile = joinpath(@__DIR__, "..", "templates", "julia_tex.tpl")
 
 function weave_file(folder,file,build_list=(:script,:html,:pdf,:github,:notebook); kwargs...)
-  tmp = joinpath(repo_directory,"tutorials",folder,file)
-  Pkg.activate(dirname(tmp))
-  Pkg.instantiate()
-  args = Dict{Symbol,String}(:folder=>folder,:file=>file)
-  if :script ∈ build_list
-    println("Building Script")
-    dir = joinpath(repo_directory,"script",folder)
-    isdir(dir) || mkpath(dir)
-    args[:doctype] = "script"
-    tangle(tmp;out_path=dir)
-  end
-  if :html ∈ build_list
-    println("Building HTML")
-    dir = joinpath(repo_directory,"html",folder)
-    isdir(dir) || mkpath(dir)
-    args[:doctype] = "html"
-    weave(tmp,doctype = "md2html",out_path=dir,args=args; fig_ext=".svg", css=cssfile, kwargs...)
-  end
-  if :pdf ∈ build_list
-    println("Building PDF")
-    dir = joinpath(repo_directory,"pdf",folder)
-    isdir(dir) || mkpath(dir)
-    args[:doctype] = "pdf"
-    try
-      weave(tmp,doctype="md2pdf",out_path=dir,args=args; template=latexfile, kwargs...)
-    catch ex
-      @warn "PDF generation failed" exception=(ex, catch_backtrace())
+    tmp = joinpath(repo_directory,"tutorials",folder,file)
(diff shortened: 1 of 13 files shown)


src/TuringTutorials.jl
```diff
@@ -7,117 +7,120 @@ cssfile = joinpath(@__DIR__, "..", "templates", "skeleton_css.css")
...
+        tangle(tmp;out_path=dir)
+    end
+    if :html ∈ build_list
+        println("Building HTML")
+        dir = joinpath(repo_directory,"html",folder)
+        isdir(dir) || mkpath(dir)
+        args[:doctype] = "html"
+        weave(tmp,doctype = "md2html",out_path=dir,args=args; fig_ext=".svg", css=cssfile, kwargs...)
+    end
+    if :pdf ∈ build_list
```
devmotion, line 28: I'm not sure if we ever want to build PDFs (and in particular add them to the repo). There are discussions over at SciML to remove them.

devmotion: You can, e.g., always print the HTML version to a file or weave it manually.

src/TuringTutorials.jl
```diff
@@ -7,117 +7,120 @@ cssfile = joinpath(@__DIR__, "..", "templates", "skeleton_css.css")
...
 end
 
 function open_notebooks()
-  Base.eval(Main, Meta.parse("import IJulia"))
-  path = joinpath(repo_directory,"notebook")
-  IJulia.notebook(;dir=path)
+    Base.eval(Main, Meta.parse("import IJulia"))
+    path = joinpath(repo_directory,"notebook")
+    IJulia.notebook(;dir=path)
 end
```
devmotion, line 124: I'm not sure if this is more convenient than just defining a function/constant that returns the notebook directory and tell users to run:
```julia
using IJulia, TuringTutorials
notebook(; dir=TuringTutorials.notebookdir())
```
That would be more explicit and avoid the nasty `eval`...

src/TuringTutorials.jl
cpfiffer: I'd actually lean towards weaving PDFs, just because academic types might prefer to print these out and use the nicer PDF typesetting.

src/TuringTutorials.jl
cpfiffer: I agree with David on this one. Having `TuringTutorials.open_notebooks()` seems like an antipattern.

src/TuringTutorials.jl
torfjelde: We could also just remove it as a default, but keep the functionality.

For starters we're likely going to have to generate everything by hand anyways, simply because the Git Actions, etc. will surely time out before completion.

src/TuringTutorials.jl
torfjelde: I agree with you guys :+1: Didn't give it any thought, but just copy-pasted.

src/TuringTutorials.jl
devmotion: I'm completely fine with keeping the functionality (although just calling `weave` doesn't seem more complicated) but I'm not sure if it's useful to generate them by default and add them to the git repo. Maybe a nice simple functions for generating PDFs could be added (instead), similar to the functionality for opening the notebook/getting the directory below?

src/TuringTutorials.jl
torfjelde: Removed it :+1:

src/TuringTutorials.jl
torfjelde: Cool! I removed PDF from the default.

torfjelde:
Is this good now?

torfjelde:
Btw, after this PR, it should just be a matter of generating the markdown, notebooks, etc., i.e. running `TuringTutorials.weave_all()`, and then checking that they all render nicely on the website (there's a way to run it locally ,right?)

That should be us ready to move to jmd instead of notebooks for tutorials :+1:

devmotion (approved):

PR #90, opened 2020-10-26
[WIP] Remove Plots from tutorials that have StatsPlots

kescobo:
[`StatsPlots` reexports `Plots`](https://github.com/JuliaPlots/StatsPlots.jl/blob/master/src/StatsPlots.jl#L8), so if you have `using StatsPlots`, then `using Plots` is redundant.

This is a relatively straightforward PR, where I've removed the `using Plots` from all of the tutorials where `using StatsPlots` is also used. There are a couple of instances where `using Plots` was used early in the tutorial, and `using StatsPlots` was used later. If you think those should stay as they were for clarity, I'm happy to revert.

I'm also a bit confused about the development process here. I edited the markdown directly, but it looks like maybe the expectation is that one would edit the notebooks, then run `render-exammple.bash`? But the README says that you have to run `weave-example.jl` if you change something, and `render-example.bash already seems to do this. Some more instruction on the steps to make this usable would be appreciated.

### DIFF
--- a/markdown/0_Introduction.md
+++ b/markdown/0_Introduction.md
@@ -24,7 +24,7 @@ First, let's load some of the packages we need to flip a coin (`Random`, `Distri
 using Random
 
 # Load a plotting library.
-using Plots
+using StatsPlots
 
 # Load the distributions library.
 using Distributions
@@ -91,9 +91,6 @@ The intuition about this definition is that the variance of the distribution wil
 
 
 ```julia
-# Import StatsPlots for animating purposes.
-using StatsPlots
-
 # Make an animation.
 animation = @gif for (i, N) in enumerate(Ns)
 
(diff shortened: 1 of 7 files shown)


devmotion:
Actually, since many tutorials are outdated the structure of the tutorials is about to change (https://github.com/TuringLang/TuringTutorials/issues/86). The process of generating the different outputs will be different as well.

@torfjelde Should we wait with merging this PR until the switch to Weave and jmd is completed?

kescobo:
Oh neat - just read through that issue - great plan! I'm happy to close this for now and come back to it once things are settled on that front, and I think that seems like the most sensible thing to do, assuming that transition is going to happen relatively soon.

That said, if it's not too much trouble, would you be willing to add a `hacktoberfest-accepted` label? :wink: If not, no worries.

cpfiffer:
Tag added.

kescobo:
Thanks! I'll close this now and watch #86, circling back when it's ready.

PR #91, opened 2020-10-26
Fix order heads and tails in the Beta distribution

rikhuijzer:
The order was reversed as could be seen by computing the posterior distribution for one heads or one tails.
For example, if you calculated the posterior for a sample of size one and containing one heads, the posterior plot would show that the parameter is most likely to be `0` whereas the plot shows that an higher value means a higher probability for heads.

Note that I haven't run the notebook again to fix the GIF. This is because it would take me about 30 minutes to setup Jupyter again. This is a minor fix, which I guess will be end up into the tutorials at some point.

### DIFF
--- a/0_Introduction.ipynb
+++ b/0_Introduction.ipynb
@@ -170,7 +170,7 @@
     "    tails = N - heads\n",
     "    \n",
     "    # Update our prior belief in closed form (this is possible because we use a conjugate prior).\n",
-    "    updated_belief = Beta(prior_belief.α + heads, prior_belief.β + tails)\n",
+    "    updated_belief = Beta(prior_belief.α + tails, prior_belief.β + heads)\n",
     "\n",
     "    # Plotting\n",
     "    plot(updated_belief, \n",


devmotion:
Isn't the (main) problem here that heads is compute as `sum(data[1:i-1])`, i.e., it will always be `0` for a sample of size 1?

rikhuijzer:
It seems that I made a mistake when copying the code. When using
```
data = [0, 1, 0, 1]
prior_belief = Beta(1, 1);
	
function plot_dist(i, N) 

    # Count the number of heads and tails.
    heads = sum(data[1:i-1])
    tails = N - heads
    
    # Update our prior belief in closed form (this is possible because we use a conjugate prior).
    updated_belief = Beta(prior_belief.α + heads, prior_belief.β + tails)

    # Plotting
    plot(updated_belief, 
        size = (500, 250), 
        title = "... after $N observations",
        xlabel = "probability of heads", 
        ylabel = "", 
        legend = nothing,
        xlim = (0,1),
        fill=0, α=0.3, w=3)
    vline!([p_true])
	end
end

plot(plot_dist(1, 0), plot_dist(2, 1), plot_dist(3, 2), plot_dist(4, 3))
```

I get

image

as expected. Apologies for the inconvenience.

PR #93, opened 2020-12-12
minor spelling mistake

jonneguyt:


cpfiffer:
This actually needs to be corrected at the notebook level [here](https://github.com/TuringLang/TuringTutorials/blob/master/2_LogisticRegression.ipynb).

PR #95, opened 2020-12-17
Fix images in DE tutorial

devmotion:
I did not realize that the links to the images are hardcoded and have to be adjusted :man_facepalming:


devmotion:
I could not find any other occurrences of `10_diffeq` in the markdown file, so hopefully everything works now. One can have a look at it at https://github.com/TuringLang/TuringTutorials/blob/bayesiandiffeq/markdown/10_BayesianDiffEq.md but the images don't work in the preview since everything is in the `markdown` folder and not in `tutorials` (I guess this is already adjusted for the webpage).

yebai:
Thanks, @devmotion - it looks good. I didn’t notice that we hardcoded image links either.

PR #97, opened 2021-01-09
Fix Markdown string interpolations

kar9222:
Fix Markdown string interpolation: `$$` to `$`


kar9222:
Sorry I just noticed the markdowns are converted from the Jupyter notebooks so the issue should be in __weave-examples.jl__ where the dollar signs are converted `txt = replace(txt, inline => raw"$$")`

```jl
function polish_latex(path::String)
    txt = open(f -> read(f, String), path)
    open(path, "w+") do f
        txt = replace(txt, raw"$$" => raw"\$\$")

        beg_align = r"\\\$\\\$\n?\\begin{align}"
        end_align = r"\\end{align}\n?\\\$\\\$?"

        txt = replace(txt, beg_align => "\$\$\n\\begin{align}")
        txt = replace(txt, end_align => "\\end{align}\n\$\$")

        # Make inline math work.
        inline = r"(?<=[^\\])\$(?!\$)"
        txt = replace(txt, inline => raw"$$")

        # If there's more than three dollar signs together, use only two.
        extra_dollars = r"\${3,}"
        txt = replace(txt, extra_dollars => raw"$$")

        write(f, txt)
    end
end
```

I never use Jupyter notebook before. Will take a look at Jupyter notebook and open another PR to fix the regex conversion in `polish_latex` above when I get a chance. Closing this.

PR #103, opened 2021-03-05
Restructure project

torfjelde:
This restructures the project so that each tutorial now has its own folder with a corresponding `Project.toml` and `Manifest.toml`. This ought to make everything completely reproducible and hopefully make it so that at least some of the TuringLang team still has hair left when we're 40.


theogf:
Is that normal that there is no `Manifest.toml` files so far?

torfjelde:
Eh so it seems like `Manifest.toml` is included in `.gitignore`. That should be changed, thanks! 

BUT I won't add any Manifest.toml files _yet_, as we need to go through them all and make sure they look good, which is going to include updating deps to most recent versions as there currently is no tutorial-local dependency tracking.

theogf (approved):

Issue #96, opened 2020-12-18
Fix link to https://github.com/TuringLang/TuringTutorials/blob/master/10_diffeq.ipynb


ChrisRackauckas:
This is a somewhat timely manner. Can this link be resurrected for just about a month? Thanks.

yebai:
Fixed by 23f06ba87bb7d6dd77d3467a75d27258194b8738. 

Let's remember to remove this in the future.

devmotion:
@ChrisRackauckas Sorry, just saw your message now. If you want to make the link persistent, you could link to a specific commit. With the switch to markdown tutorials, most links to files in the master branch will become invalid, I assume.

ChrisRackauckas:
It's fine. It's just on the day of a paper release I don't want a broken link since Twitter doesn't allow editing. In a month we can remove this, but for now it's going to get a ton of traffic and "doing it well" is probably better than "doing it right"

devmotion:
I see :slightly_smiling_face: I'm sorry about the inconvenience it caused, I didn't know that you had distributed a link to the repo when fixing the webpage.

ChrisRackauckas:
No worries, it's all good in the end. That's just my impatience to make sure the paper hit the optimal release timing.

torfjelde:
Is it okay if we delete the file now? We're moving to a new format which doesn't use Jupyter notebooks, and so ideally we'd not have to keep the old ones around. We'll also add in redirects, so in the future always link to the website and the underlying notebook can be found from there:)

ChrisRackauckas:
yes, go for it.

PR #107, opened 2021-03-07
Added variational inference and differential equations markdown

torfjelde:
Final markdown generated for the tutorials generated.


devmotion:
This seems to be the outdated first iteration of the diffeq tutorial? The most recent version should contain sections about missing data and DDEs as well: https://turing.ml/dev/tutorials/10-bayesiandiffeq/

torfjelde:
Ah yes, so this is because master has a couple of commits ahead of `jmd-staging`. I think we just gotta apply these by hand at this point.

torfjelde:
But can we get this PR merged and then I'll update the bayes diff eq tutorial in a separate PR.

devmotion:
We should just be careful with this specific tutorial - when it was updated, it was messed up and incorrectly moved to a different location, and therefore not shown on the webpage. This was discovered at some point and fixed by moving the file back to the old location, but this then invalidated some links in a publication, so it was fixed by copying it back to the other location (there's sn issue about it). So currently two (hopefully equivalent) files exist in the repo.

torfjelde:
Can you use symlinks? Does that work?

devmotion:
Certainly relative symlinks would work in the repo but I am not sure if the script for building the webpage can deal with symlinks. However, I think it might be better to just remote the duplicate (i.e. "10_diffeq") again - in the [issue](https://github.com/TuringLang/TuringTutorials/issues/96) Chris mentioned that it could be removed after a month or so.

torfjelde:
If `open` follows symlinks (which I would presume it does?), then I think this would actually work very nicely. With the new branch which is compatible with the new TuringTutorials, it will walk TuringTutorials, open file, read, and then write to new file. I do this to circumvent permission issues. As a result, TuringTutorials could simply contain a symlink and the resulting structure in the website would be an "materialized" version of the link.

torfjelde:
But I still think it should be done server-side rather than in TuringTutorials, e.g. using https://github.com/jekyll/jekyll-redirect-from. 
And regarding having a link to the repository (ref the issue above), wouldn't this be fixed by simly linking to a specific commit?

torfjelde:
https://github.com/TuringLang/TuringTutorials/issues/110

devmotion:
I don't know, I still think that 10_diffeq.ipynb should just be deleted as it was requested to be up for only a month in December. But I assume redirects are needed anyway to not break links to the webpage after the transition to the new format?

torfjelde:
Yeah exactly: delete file because >1 month but need redirects for the website due to changing the naming for the new format (also because we're for sure going to do renaming, etc. in the future, so just having a way to redirect is nice as this ensures that linking to the website is always consistent).

PR #109, opened 2021-03-07
Updated DiffEq tutorial according to master

torfjelde:
Currently master is a couple of commits ahead of `jmd-staging` which include changes to the DiffEq tutorial; this PR adds those changes to the jmd branch.

I'm currently trying to regenerate the outputted markdown file but it's taking aaaages.


devmotion:
As mentioned in https://github.com/TuringLang/TuringTutorials/pull/107#issuecomment-792244563, IIRC there exist two files for this tutorial in the repo and one has to be a bit careful.

torfjelde:
Btw, I'll add the output markdown for the last tutorial to this PR in like 30mins (when it's finished), and then this is ready.

torfjelde:
Btw, if you actually want to look at the resulting tutorials, just check out the `markdown/` directory on this branch.

Issue #116, opened 2021-04-24
Add Manifest.toml and CompatHelper


devmotion:
Currently, only the Project.toml files are checked in, it seems. This helps to avoid conflicts between dependencies of different tutorials but it does not ensure reproducibility. It would be easier to rerun and debug the examples if the Manifest.toml would be checked in as well.

Additionally, one could then use CompatHelper to be informed about updated packages (it would be cool if the tutorials would be rebuilt automatically but it is even helpful if this has to be done manually, I think).

rikhuijzer:
> it would be cool if the tutorials would be rebuilt automatically but it is even helpful if this has to be done manually, I think

@cpfiffer Is there a particular reason that the docs are not built with CI? It frightens me a bit when I look at a tutorial and see "Julia Version 1.5.3", which means it's likely built months ago.

rikhuijzer:
> It would be easier to rerun and debug the examples if the Manifest.toml would be checked in as well.

In my experience with my blog (https://github.com/rikhuijzer/huijzer.xyz) and TuringModels.jl (https://github.com/StatisticalRethinkingJulia/TuringModels.jl), just the `Project.toml` with compat entries should be enough if you rerun everything about every week. I've never gotten into major issues because package maintainers stick to semantic versioning quite well.

devmotion:
It's a bit wasteful to rerun everything every week, and it's also not guaranteed that package updates (even according to Semver) don't break tutorials (e.g. additional exports in some package might cause ambiguity errors). IMO both Manifest and Project files should be added to make it reproducible, and we should use the same CI setup as SciMLTutorials and SciMLBenchmarks (https://github.com/TuringLang/TuringTutorials/pull/118#issuecomment-839730947).

rikhuijzer:
> It's a bit wasteful to rerun everything every week, and it's also not guaranteed that package updates (even according to Semver) don't break tutorials.

With that logic, we could also stop using CI. I would argue that the time wasted by people on manually running tutorials and messing around with broken tutorials is more wasteful. But, do these tutorials take hours to run? Maybe that's what I'm missing here.

[...]

PR #118, opened 2021-05-12
Fix typo in SDE fitting

ChrisRackauckas:


ChrisRackauckas:
How do you make it spawn a rebuild job?

devmotion:
There is no rebuild job, everything is built manually (and IIRC there is also some delay until it ends up on the website but maybe that changed) 😢

ChrisRackauckas:
Hmm, maybe we should just get you setup with the new SciMLBenchmarks setup.

devmotion:
I guess that maybe even a simpler setup could be sufficient since most tutorials are much less compute intensive than the benchmarks in SciML. In general, the whole jmd setup is still quite new, e.g., Manifest files and a CompatHelper setup are missing (https://github.com/TuringLang/TuringTutorials/issues/116).

ChrisRackauckas:
Being less compute intensive would just mean it would make sense to use a different queue, i.e. use the non-exclusive AMDCI

devmotion:
It would be great if it is possible to use the same setup as in SciMLBenchmarks, I just assumed the benchmarks run on some private workstations and clusters and it might be easier with some simple CI setup initially.

ChrisRackauckas:
It just uses the Julialang Buildkite which we have a ton of extra servers on there for long jobs. @staticfloat we need to get SciMLTutorials and FluxBench using it too, so we should talk about how to start distributing this around the ecosystem.

PR #123, opened 2021-07-17
Add CI

rikhuijzer:
This PR sets up CI for building the tutorials. The suggested CI job consists of the following main steps:

1. Build the tutorials* in parallel, see the section **CI output** below for an example of how the CI logs will look. This build takes about 60 to 90 minutes. Hopefully, we can speed this up a bit more in the future which will benefit users and developers. Only when a tutorial run fails, the logs are shown in the CI log. Note that, like with tests, the job fails _after_ running the code and not during.

2. Write the output to a branch called `artifacts`.

To get the output to show up on https://turing.ml, we only need to change 
```
Pkg.develop(url="https://github.com/TuringLang/TuringTutorials")
```
to
```
Pkg.add(url="https://github.com/TuringLang/TuringTutorials", rev="artifacts-")
```
in [turing.ml/make-utils.jl](https://github.com/TuringLang/turing.ml/blob/6a7e075eef36ef344700064718b6fb14333f4bce/make-utils.jl#L1-L2). Note that I haven't removed the artifacts from `TuringTutorials#master`, so we can always switch this line back if the CI job doesn't work for some reason.

\* @cpfiffer told me that tutorial 10 (differential equations) takes ≈10 hours to complete. Therefore, this tutorial has been excluded from CI for now. My hope is that we can speed the tutorial up and add it later.

- Closes #116
- Closes #101
- Closes #74

## Demo

A job has finished in my fork of TuringTutorials: https://github.com/rikhuijzer/TuringTutorials/actions/runs/1040205250.

The output of the run can be inspected at https://github.com/rikhuijzer/TuringTutorials/tree/artifacts-ci. To verify that the CI really updated something, compare https://turing.ml/dev/tutorials/00-introduction/ which shows Julia Version **1.5.3** with [00_introduction.md](https://github.com/rikhuijzer/TuringTutorials/blob/artifacts-ci/markdown/00-introduction/00_introduction.md) which shows Julia Version **1.6.2**.

## A note on tutorial 3

In the run, tutorial 3 (neural networks) was failing on:
```
advi = ADVI(10, 5_000)
q_hat = vi(m, advi, q);
```
with a `BoundsError`. This is a known issue: #59. I've removed the part for now.

## CI output

```
00-introduction: Starting build
05-linear-regression: Starting build
00-introduction: Build took 7.3 minutes and allocated 18.98 GiB:
00-introduction: Writing log to /home/runner/work/TuringTutorials/TuringTutorials/tutorials/00-introduction/weave.log
[...]

PR #124, opened 2021-07-17
Fix typo in folder name

rikhuijzer:
Patch for the failing CI on `master`. Was introduced by #123.


rikhuijzer:
Although this now runs successfully, `master` shouldn‘t have failed. I‘ll further look into this tomorrow.

rikhuijzer:
Ok, for some reason tutorial 8 (multionomial-regression) failed on `master` and the stacktrace wasn't visible because one of the folder names didn't match the `jmd` name. That last issue is now fixed. 

The failure of tutorial 8 is not fixed, because I cannot reproduce it. However, the next time that it happens, there will at least be a stacktrace thanks to this PR.

PR #126, opened 2021-07-18
Use MCMCThreads in tutorial 1, 2 and 7

rikhuijzer:
I was already wondering why tutorial 1, 2 and 7 were so surprisingly slow.


yebai (approved):
Looks good to me, thanks for the improvements @rikhuijzer!

PR #128, opened 2021-07-21
Only weave changed tutorials

### LINKED ISSUE #127: Do not weave unchanged tutorials
Currently, CI reruns all tutorials it seems. Usually this is not needed and it would be sufficient to only build tutorials whose jmd file or Project or Manifest file were changed (as done eg in https://github.com/SciML/SciMLTutorials.jl). This would be a lot faster than the current setup.

rikhuijzer:
Thanks to @devmotion for the very nice idea behind this PR. It was pretty easy to implement and saves super much time.

I've decided to implement the cache in Julia instead of Buildkite. (They use Buildkite at https://github.com/SciML/SciMLTutorials.jl.) This has two benefits, namely:

1. The logic is less dependent on the continuous integration provider. In other words, with a mostly Julia implementation, it's easy to switch from GitHub Actions to any other provider.
1. Developers are unlikely to be familiar with Buildkite.

Feel free to comment on code that I've added here, or on things that I've added in #123.

- Closes #127.

### DIFF
--- a/src/TuringTutorials.jl
+++ b/src/TuringTutorials.jl
@@ -9,7 +9,7 @@ using Plots
 using Requires
 using Weave
 
-const repo_dir = string(pkgdir(TuringTutorials))::String
+const REPO_DIR = string(pkgdir(TuringTutorials))::String
 
 include("cache.jl")
 include("build.jl")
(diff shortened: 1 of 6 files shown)


src/TuringTutorials.jl
```diff
@@ -9,12 +9,16 @@ using Plots
 using Requires
 using Weave
 
-export build_folder, build_all, verify_logs, tutorials
+const repo_dir = string(pkgdir(TuringTutorials))::String
```
cpfiffer, line 12: This should probably be `REPO_DIR` instead of `repo_dir` to clarify that it's a constant.

cpfiffer (approved):

PR #221, opened 2021-08-01
Avoid hidden state in artifacts branch

rikhuijzer:
Something is wrong with the artifacts branch since the caching was introduced. The output is the old one again because it is build with Julia 1.5. See, for example, https://github.com/TuringLang/TuringTutorials/blob/artifacts/markdown/06-infinite-mixture-model/06_infinite-mixture-model.md. 

To solve this and some other problems with the output, the code now explicitly copies all the files it needs and puts everything as one orphaned commit into the branch. Now, the output will either work or it won't, because the hidden state is avoided. This makes verifying correctness and debugging much easier.

I guess that one day the CI functionality will work flawlessly :stuck_out_tongue:. Till that time, I'll keep sending in PRs :stuck_out_tongue_closed_eyes:.

**This PR needs a thorough review. I'm not so sharp anymore today**


cpfiffer:
Aight let's give it a shot!

rikhuijzer:
Thanks @cpfiffer! It broke, I see. I'll look at it tomorrow :+1: It didn't push anything, so it's not a priority to fix it now.

PR #249, opened 2021-08-16
Update tutorial 09 variational inference

rikhuijzer:
- Add assertions.
- Closes #232.
- Closes #233.
- Closes #234.
- Closes #235.


rikhuijzer:
@torfjelde, quick question. You have worked on tutorial 9 on variational inference. When updating the tutorial, I get 1.0092761373397834 instead of 1.023994570402229 (see https://turing.ml/dev/tutorials/09-variational-inference/). 

In summary, with older dependencies the assertion 
```
v, m = (mean(rand(q, 1000); dims = 2)..., )                                               
@assert isapprox(v, 1.022; atol=0.01)
```
holds, but with newer versions, I should change atol to 0.1 to make it pass. Would you say that that this more relaxed atol is also okay, or was the high precision essential for the tutorial?

rikhuijzer:
I guess that it's better to have a slightly less accurate tutorial than having a tutorial with outdated dependencies. Readers want to use the latest version and not some version from last year. I'm currently updating the Bijectors and Turing dependencies and seeing whether that works.

rikhuijzer:
Below, it's visible that the atol for the variance increased from 0.01 to 0.1. The difference is caused by updating Turing from 0.14 to 0.15 or newer.

Because readers probably rather have slightly less accurate tutorials than tutorials with outdated versions, I'd say this is acceptable.


## Before this PR
```
var(x), mean(x)
```
```
(1.0225001600719719, -0.027900450605557185)
```
```
(mean(rand(q, 1000); dims = 2)..., )
```
```
(1.023994570402229, -0.026903362671990064)
```

## After this PR
```
var(x), mean(x)
```
```
(1.0225001600719719, -0.027900450605557185)
```
```
(mean(rand(q, 1000); dims = 2)..., )
```
```
(1.0092761373397834, -0.0280045068429487)
```

rikhuijzer:
On the upside, the loss on the test set got even smaller (very slightly, but still).

@torfjelde, can you review and merge if you agree?

torfjelde (approved):
Wonderful @rikhuijzer ! Thanks:)

Issue #218, opened 2021-07-24
Ease review by using assertions


rikhuijzer:
The post on https://discourse.julialang.org/t/package-growth-tips-and-tricks-creating-an-organization-triage-better-management/65167/2 got me thinking about the current setup of TuringTutorials.

The tutorials are now very fragile because:

- the tutorials aren't tested and
- the tutorial build logic isn't tested.

For example, Weave could change in such a way that errors aren't correctly recognized by the build. This makes reviewing in the train impossible (see the aforementioned post). 

The solution would be to add more tests for the build logic and add assertions in the tutorials.

I'm assigning this issue to myself, but others are, of course, also welcome to pick it up.

- [x] Add tests to build logic
- [x] Add test to verify that build fails if assertion doesn't hold
- [x] Add assertions to tutorial 00 - introduction (#278)
- [x] Add assertions to tutorial 01 - gaussian mixture model (#276)
- [x] Add assertions to tutorial 02 - logistic regression (#248)
- [ ] Add assertions to tutorial 03 - bayesian neural network
- [ ] Add assertions to tutorial 04 - hidden markov model
- [x] Add assertions to tutorial 05 - linear regression (#277)
- [ ] Add assertions to tutorial 06 - infinite mixture model
- [ ] Add assertions to tutorial 07 - poisson regression
- [ ] Add assertions to tutorial 08 - multinomial logistic regression
- [x] Add assertions to tutorial 09 - variational inference (#249)
- [ ] Add assertions to tutorial 10 - bayesian differential equations

PR #269, opened 2021-08-19
Update CompatHelper.yml

rikhuijzer:
CompatHelper is opening too many PRs due to some internal rewrites on their side. Let's disable it for a few months and try to get the already open PRs done. We don't need extra PRs for now.

See also: https://github.com/JuliaRegistries/CompatHelper.jl/issues/378 or other issues at CompatHelper.

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -16,16 +16,4 @@ jobs:
         run: |
           using Pkg
           Pkg.add("CompatHelper")
-
-      - name: Run CompatHelper
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          COMPATHELPER_PRIV: ${{ secrets.COMPATHELPER_PRIV }}
-        shell: julia --color=yes {0}
-        run: |
-          using CompatHelper
-          dir = "tutorials"
-          tutorials_dirs = readdir(dir)
-          subdirs = joinpath.(dir, tutorials_dirs)
-          subdirs = [""; subdirs]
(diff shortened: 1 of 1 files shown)


devmotion:
An alternative (better?) approach would be to use CompatHelper 2 if CompstHelper 3 causes problems.

devmotion:
Just a heads up: I'm not completely sure what's the common practice in TuringTutorials but in most other Turing repos we follow [ColPrac](http://colprac.sciml.ai/) which requires approval of another contributor with merge rights before a PR is merged.

rikhuijzer:
Yes, you're right. Usually, I do wait for approval, but this morning I got very annoyed by the CompatHelper PRs because I've worked hours and hours to get rid of those PRs.

Next time, I'll wait for review again.

yebai:
> Yes, you're right. Usually, I do wait for approval, but this morning I got very annoyed by the CompatHelper PRs because I've worked hours and hours to get rid of those PRs.

I've closed these Compat PRs in batch. We can re-enable CompatHelper (if needed).

rikhuijzer:
> I've closed these Compat PRs in batch. We can re-enable CompatHelper (if needed).

As long as the compat entries aren't updated, this would re-open them again next week if I'm not mistaken.

The only real solution here is to update the compat entries, which I'm working on (for example, #248 and #249). My idea was to keep CompatHelper disabled for a bit and first fix the existing backlog of CompatHelper PRs.

rikhuijzer:
The progress of adding assertions can be tracked here: https://github.com/TuringLang/TuringTutorials/issues/218.

PR #278, opened 2021-09-11
Add assertion to tutorial 0 and workaround for thread-unsafe Pkg.jl

rikhuijzer:
This PR adds an assertion (as part of #218).

Also:

- Remove whitespace
- Use `@model function foo` instead of `@model foo = begin`.
- Update dependencies (related to #223 and #250).
- Add workaround for thread-unsafe Pkg.jl.


src/build.jl
```diff
@@ -65,6 +65,20 @@ function build_folder(folder)
     write(path, log)
 end
 
+"""
+    safe_instantiate(folders)
+
+Install all packages sequentially since Pkg.jl is thread-unsafe.
+See https://github.com/JuliaLang/Pkg.jl/issues/2219 for details.
+"""
+function safe_instantiate(folders)
```
devmotion, line 74: The problem with this approach is that you end up with the last activated environment when you run `safe_instantiate` or `parallel_build`, i.e., it will affect your project environment when you run it from the REPL. Additionally, in my experience it is quite messy and easily leads to broken/incorrectly loaded environments if one activates and instantiates multiple environments in one session (but maybe it's less of an issue if no packages are loaded, I am not sure how problematic precompilation is).

Instead one could use the same approach as in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/671f960726b0b9826b9e5a81cc3e55d6c4ba0912/docs/make.jl#L10-L34 (and similarly in AbstractGPs). BTW why does TuringTutorials use multiple threads instead of running multiple system commands asynchronously? Are there any advantages?

rikhuijzer: > The problem with this approach is that you end up with the last activated environment when you run safe_instantiate or parallel_build, i.e., it will affect your project environment when you run it from the REPL. Additionally, in my experience it is quite messy and easily leads to broken/incorrectly loaded environments if one activates and instantiates multiple environments in one session (but maybe it's less of an issue if no packages are loaded, I am not sure how problematic precompilation is).

Ah okay. Yes. Good point, that could indeed lead to issues. I could start a new Julia instance for each activate and instantiate. Would you consider that as a solution?

> Instead one could use the same approach as in https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/blob/671f960726b0b9826b9e5a81cc3e55d6c4ba0912/docs/make.jl#L10-L34 (and similarly in AbstractGPs). BTW why does TuringTutorials use multiple threads instead of running multiple system commands asynchronously? Are there any advantages?

In essence, we're doing the same thing, right? I just copied the `$(Base.julia_cmd()) ...` from there. The reason that I use threads is because it explicitly states (to the reader) that it runs in parallel. With `map` and `process`, it's a bit more implicit.

devmotion: >  I could start a new Julia instance for each activate and instantiate. Would you consider that as a solution?

That's what we do in the code I linked 🙂 It seems to work so I guess it should be fine.

rikhuijzer: :+1: Now I see.

devmotion: > I just copied the $(Base.julia_cmd()) ... from there. The reason that I use threads is because it explicitly states (to the reader) that it runs in parallel. With map and process, it's a bit more implicit.

But it requires the user to set JULIA_NUM_THREADS, otherwise it won't run in parallel, will it? Whereas in the other case it just starts multiple system process in parallel and hence the system scheduler will take care of everything.

rikhuijzer: Aha, okay. And then the new Julia instances use multiple threads because the OS gives them a new thread if needed. So, you spawn all the Julia instances at the same time? Will this not actually slow down the execution when you start many tutorials. In the case of TuringTutorials, we have 10 tutorials and some of them start `n` threads. On the GitHub CI, I set `JULIA_NUM_THREADS` to 2, so now we end up with at most 4 threads running at the same time (with the GitHub runners having 2 threads). If I would use map and spawn all tutorials at the same time, we would end up with somewhere between 10 and 20 threads (10 tutorials * 1.5 threads per tutorial on average).

Could you open an issue if you want TuringTutorials to remove the `Threads.@threads`? I think that it is out of scope for this PR, because it needs some proper testing when I change it.

rikhuijzer:
The safe instantiate now uses multiple processes.

I've also deleted some old files. I was planning to do that later in a separate PR, but I accidentally committed some of these files. Let me know if I should revert that.

script/01-gaussian-mixture-model/01_gaussian-mixture-model.jl
```diff
@@ -23,46 +23,56 @@ using Turing, MCMCChains
...
     # Comment out this line if you instead want to draw the weights.
     w = [0.5, 0.5]
-    
+
     # Draw assignments for each datum and generate it from a multivariate normal.
     k = Vector{Int}(undef, N)
     for i in 1:N
         k[i] ~ Categorical(w)
-        x[:,i] ~ MvNormal([μ[k[i]], μ[k[i]]], 1.)
+        x[:,i] ~ MvNormal([μ[k[i]], μ[k[i]]], 1.)
```
devmotion, line 51: This MvNormal constructor is deprecated.

rikhuijzer: Aha. I see. I also see this problem in tutorial:

- 03 Bayesian neural network
- 05 Linear regression
- 08 Multinomial logistic regression
- 09 Variational inference
- 10 Bayesian differential equations

I'll open a separate issue for this.

rikhuijzer:
@devmotion, I've enabled --depwarn=yes now in the CI which might avoid committing deprecated tutorials in the future. I've also enabled --depwarn=yes now on my pc by default.

Can we merge this PR? I've opened an issue for the deprecation and plan to pick that one up soon in anther PR.

devmotion:
The changes to the Gaussian mixture model files are not mentioned above and seem unrelated?

devmotion:
> @devmotion, I've enabled --depwarn=yes now in the CI which might avoid committing deprecated tutorials in the future. I've also enabled --depwarn=yes now on my pc by default.
> 
> Can we merge this PR? I've opened an issue for the deprecation and plan to pick that one up soon in anther PR.

I don't think we should do it, I would suggest reverting these changes. It is completely fine to use and run the deprecated code for users (and us). The main point of these deprecations is that the code still works, it's just not future proof and we might have to update it when updating to an upcoming breaking release of Distributions. But then we'll notice that it has to be updated since it will be broken. Enabling deprecation warnings just increases build times (potentially even causes timeouts, we've observed this in Turing) and creates noisy outputs - and we still won't notice them if we do not carefully inspect the logs or outputs.

That being said, I think if we update or add tutorials it would be strange to use the deprecated version if we are aware of the deprecations since we have to update them at some point eventually (since I introduced these deprecations I noticed the deprecated implementation 😂).

rikhuijzer:
> The changes to the Gaussian mixture model files are not mentioned above and seem unrelated?

See above, where I said: _"I've also deleted some old files. I was planning to do that later in a separate PR, but I accidentally committed some of these files. Let me know if I should revert that."_

> Enabling deprecation warnings just increases build times (potentially even causes timeouts, we've observed this in Turing) and creates noisy outputs - and we still won't notice them if we do not carefully inspect the logs or outputs.

Good to know. It seems to have already happened at 4263aab. I canceled that run.

> That being said, I think if we update or add tutorials it would be strange to use the deprecated version if we are aware of the deprecations since we have to update them at some point eventually

Agreed. I'll keep https://github.com/TuringLang/TuringTutorials/issues/279 open then.

devmotion (approved):
> See above, where I said:

Ah I missed this completely 😬

rikhuijzer:
> Ah I missed this completely :grimacing:

No problem. Thanks for the review :+1:

PR #280, opened 2021-09-24
Move to buildkite build system

devmotion:
This PR is a draft of the buildkite integration. Most code is copied from or modified according to https://github.com/SciML/SciMLTutorials.jl. Probably something is broken currently :stuck_out_tongue: 

The only problem is that https://github.com/staticfloat/coppermind-buildkite-plugin (used for memoization and avoiding unnecessary builds) seems to require a custom S3 bucket (e.g., SciMLTutorials uses s3://julialang-buildkit-artifacts/scimltutorials) and to not work with the default buildkite artifact system since one has to provide credentials explicitly. Is this correct, @staticfloat? And do we have to set up our own S3 bucket or is it possible to use something like s3://julialang-buildkit-artificats/turingtutorials?


.buildkite/publish_tutorials_output.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+# Ensure that our git wants to talk to github without prompting
+ssh-keyscan github.com >> ~/.ssh/known_hosts
+
+# Clone TuringTutorials to temporary directory
+temp_dir=$(mktemp -d)
+git -C "${temp_dir}" clone --branch artifacts git@github.com:TuringLang/TuringTutorials .
```
devmotion, line 8: I guess a possibly cleaner and more public (and hence easier discoverable?) alternative would be to push to a separate repo such as TuringTutorialsOutput (SciMLTutorials pushes to SciMLTutorialsOutput).

staticfloat:
>  Is this correct, @staticfloat?

Yes, that's correct.  You can't use buildkite artifacts, because those artifacts are not easily available to other jobs.

> is it possible to use something like s3://julialang-buildkit-artificats/turingtutorials?

Do you guys already have an S3 account?  It's not too hard to set up, and will probably be quite cheap.  We have our bucket configured to auto-delete things after 30 days, which helps us not pay ever-increasing amounts of money for a build cache.

src/TuringTutorials.jl
```diff
@@ -1,146 +1,98 @@
...
+        Pkg.activate(folder)
+        Pkg.instantiate()
+        Pkg.build()
+    end
+
     args = Dict{Symbol,String}(:folder => folder, :file => file)
-    if :script ∈ build_list
-        println("Building Script")
-        dir = joinpath(REPO_DIR, "script", folder)
-        isdir(dir) || mkpath(dir)
```
rikhuijzer, line 59: nice. well spotted

rikhuijzer (commented):
I have no intuition for the buildkite system, so also no idea whether it will work.

yebai:
@staticfloat Turing doesn’t have any S3 setup yet. It would be helpful to share a Julia’s S3 bucket if possible. I don’t expect Turing tutorials will consume a lot of spaces.

.buildkite/run_tutorial.yml
```diff
@@ -0,0 +1,73 @@
...
+            # We are sensitive to the source code of this package changing
+            - src/**/*.jl
+            # We are sensitive to our overall dependencies changing
+            - ./*.toml
+          outputs:
+            - html/**/*.html
+            - markdown/**/*.md
+            - notebook/**/*.ipynb
+            - pdf/**/*.pdf
+            - script/**/*.jl
```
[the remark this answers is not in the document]
devmotion: I'm not sure, this is something we would have to check and possibly adapt when the S3 bucket is available. The images are also available separately: https://github.com/SciML/SciMLTutorialsOutput/tree/master/markdown/advanced/figures

yebai:
Thanks @staticfloat - that sounds great!

devmotion:
Any progress here?

devmotion:
@staticfloat what's the path forward here? Should I just update this PR with the recent changes in SciMLTutorials? Or is there more we have to fix regarding the S3 bucket?

devmotion:
> Yeah, just copy over all the SciMLTutorials stuff. We'll just have you use the same bucket for now. :)

OK, I copied most files and replaced `SciML` with `Turing` - should I copy the `.key` and `.signature` files as well? I guess also one of the BuildKite admins has to enable the pipeline?

devmotion:
It seems the buildkite run passed, so it seems at least something is working 🙂 But I can't look at the details (seems I need a buildkite login?), is this expected?

staticfloat:
Heh, I forgot to make the pipeline public.  :P

The way the S3 upload credentials are managed is via the `cryptic` plugin, which uses cryptographic signatures to verify that pieces of the build pipeline have not been modified (e.g. so that someone can't come by and make a PR that says `echo ${AWS_SECRET_KEY}`.

As you will now be able to see, the build currently pauses when attempting to run the S3 upload step, because the signatures don't match.  This is because some of the files that were being hashed ([these ones](https://github.com/TuringLang/TuringTutorials/pull/280/files#diff-82f5dfc4b826c73034089dbf1f60b934cf7d72b0688531e748782691e5ef577eR13-R15)) were modified when you brought them into this repository.  The pipeline will automatically pause and allow a buildkite admin to override the hash check by clicking a button in the buildkite webui.  I suggest creating a buildkite account, then sending me your account name/email so that I can add you as an admin for this repository, and you can unblock pipelines when you need to make a change and I'm not here to update the signatures.  Updating the signatures shouldn't be needed often (only when we change how things get uploaded to AWS).

staticfloat:
Signature update PR here: https://github.com/TuringLang/TuringTutorials/pull/283 (can't push directly to this branch)

staticfloat:
Great; so we can now see that the signatures allow execution just fine, but there was a legitimate test failure.

staticfloat:
I've invited you to join the buildkite group!

devmotion:
Thanks, I've accepted the invite!

I copied some additional things and cleaned the repo a bit, tests seem to pass now and the test artifacts look reasonable.

devmotion:
It looks quite promising, the tutorials seem to build after updating the manifest files (no breaking changes of direct dependencies such as Turing etc.): https://buildkite.com/julialang/turingtutorials/builds/12 Only the DiffEq tutorial is broken it seems (but probably better to fix this separately).

devmotion:
>  Updating the signatures shouldn't be needed often (only when we change how things get uploaded to AWS).

I changed `launch_tutorials.yml`, so the signature does not match anymore and I had to bypass the check manually in the web interface :grimacing: The `figures` directory with the figures generated by Markdown execution was missing from the artifacts (it seems in SciMLTutorials Chris added e.g. https://github.com/SciML/SciMLTutorialsOutput/tree/master/markdown/advanced/figures manually :thinking:). It seems to work: https://buildkite.com/julialang/turingtutorials/builds/14#7e9f5599-3957-40d9-bb71-376062b7321a

devmotion:
I think this PR is ready for a proper review. You can check the buildkite results and have a look at the artifacts.

Project.toml
```diff
@@ -3,19 +3,20 @@ uuid = "09eb8af7-3c66-4d0b-a457-e0c10c662b2b"
...
-Plots = "1.19"
 Requires = "1.1"
 Weave = "0.10"
 julia = "1.5"
+
+[extras]
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+
+[targets]
+test = ["Test"]
```
rikhuijzer, line 22: Why this Julia <= 1.2 way of managing test dependencies?

src/TuringTutorials.jl
```diff
@@ -1,186 +1,134 @@
 module TuringTutorials
 
-import IOCapture
+import Requires
```
rikhuijzer, line 3:
```suggestion
using Requires: @require
```
Makes it more clear that we don't extend anything from `Requires`.

src/TuringTutorials.jl
```diff
@@ -1,186 +1,134 @@
...
-function polish_latex(path::String)
-    # TODO: Is it maybe better to overload https://github.com/JunoLab/Weave.jl/blob/b5ba227e757520f389a6d6e0f2cacb731eab8b12/src/WeaveMarkdown/markdown.jl#L10-L17
-    # and replace the `tex.formula` there? Only negative part is that this of course
-    # will affect all of the markdown parsers, which is not necessarily desirable.
-    txt = open(f -> read(f, String), path)
-    open(path, "w+") do f
-        txt = replace(txt, "\\\\\n" => "\\\\\\\\\n")
-        write(f, txt)
-    end
+    Requires.@require Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80" include("weaveplots.jl")
```
rikhuijzer, line 17:
```suggestion
    @require Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80" include("weaveplots.jl")
```
Together with previous suggestion.

src/TuringTutorials.jl
```diff
@@ -1,186 +1,134 @@
...
+function weave(folder::AbstractString, file::AbstractString; out_path_root::AbstractString=pwd(), build::Tuple{Vararg{Symbol}}=DEFAULT_BUILD)
+  if !issubset(build, (:script, :html, :pdf, :github, :notebook)) 
+    throw(ArgumentError("only build types :script, :html, :pdf, :github, and :notebook are supported"))
+  end
+
+  target = joinpath(REPO_DIR, "tutorials", folder, file)
+  @info("Weaving $(target)")
+  
+  # Activate project
+  # TODO: use separate Julia process?
```
rikhuijzer, line 29: Really? Is this not managed by the buildkite stuff? If it ever gives problems please consider switching to PlutoStaticHTML. There, separate Julia processes and all that kind of stuff is managed by Pluto.

src/weaveplots.jl
```diff
@@ -1,5 +1,8 @@
 # HACK: So Weave.jl has a submodule `WeavePlots` which is loaded using Requires.jl if Plots.jl is available.
 # This means that if we want to overload methods in that submodule we need to wait until `Plots.jl` has been loaded.
+import .Plots
+import .Weave
```
rikhuijzer, line 4: This plotting kind of stuff with grabbing outputs and stuff is part of PlutoStaticHTML. I don't know whether PlutoStaticHTML can manage GIFs though. Pluto can, I think

tutorials/02-logistic-regression/02_logistic-regression.jmd
```diff
@@ -233,7 +233,7 @@ println("Not defaults: $not_defaults
     Percentage non-defaults correct $(predicted_not_defaults/not_defaults)")
 ```
 
-```julia; echo=false
+```julia; echo=false; error=false
```
rikhuijzer, line 236: Why were these assertions disabled again? The point of them is to verify the output of the notebook

rikhuijzer (approved):
Thanks for your work here David. I dislike a bit about this that we still need to handle a lot of Plots specific logic and some Buildkit stuff which almost no one knows, but overall this PR looks like a major improvement so I'd say let's go :smiley: I'm sure that any further problems if arising can be fixed quite easily :ok_hand:

src/TuringTutorials.jl
staticfloat: I can speak to the buildkite stuff: the way this repository is set up, each tutorial `.jmd` file that gets modified will be a separate build job in buildkite; [you can see a list of build jobs](https://buildkite.com/julialang/turingtutorials/builds/17) like `tutorials/05-linear-regression` and `tutorials/04-hidden-markov-models`, etc...  Each one of those jobs is a separate Julia process (indeed, they may even be running on separate machines) so I'm not 100% certain what the intent behind this comment is.

src/TuringTutorials.jl
devmotion: `import Requires` is not sufficient to extend any method or macro from Requires directly. It only loads `Requires` in the namespace and is equivalent with `using Requires: Requires`.

src/TuringTutorials.jl
devmotion: Yes, buildkite is fine and not the reason for the comment. The motivation is users that weave a single or multiple tutorials. Since it only affects pdf, html, and markdown output I don't expect it to be a major issue though.

tutorials/02-logistic-regression/02_logistic-regression.jmd
devmotion: They are enabled. This option ensures that `weave` errors. Otherwise errors only end up in the output.

Project.toml
devmotion: Since it works and is simpler than the new, still not completely stable, test/Project.toml way (see e.g. https://github.com/JuliaDiff/ChainRules.jl/issues/574#issuecomment-1016565103).

cpfiffer (approved):
LGTM, I'm happy to try to get this up and running on `main`.

src/TuringTutorials.jl
rikhuijzer: > `import Requires` is not sufficient to extend any method or macro from Requires directly.

No, but it makes it possible and that's why many people stick to explicit loading of symbols via `using X: y`.

devmotion: No, it doesn't. You still have to write `Requires.f(...) = ...` if you want to extend `Requires.f`.

It would be different if you use `import Requires: f`. Then you could extend it with `f(...) = ...`.

rikhuijzer: Irregardless of what's best here, this shouldn't stand in the way of merging.

Project.toml
rikhuijzer: Hmm nice yes that's a fair critique there.

yebai (approved):
Great work - many thanks @devmotion!

devmotion:
@staticfloat Do you know what the problem is on the master branch? It didn't run so I triggered buildkite manually but it fails with an error that points to the pre-command hook of coppermind: https://buildkite.com/julialang/turingtutorials/builds/18#5c8ac6f4-aad8-473c-a5fe-c369e5432a10

staticfloat:
It was a worker misconfiguration; I fixed it and re-ran your job here: https://buildkite.com/julialang/turingtutorials/builds/22#19e0274a-5ad5-4426-8eaa-a792be86b491

Looks like the publishing step has some problems; we're trying to upload with an encrypted deploy key that is supposedly associated with https://github.com/TuringLang/TuringTutorialsOutput but I bet it's actually the encrypted deploy key for the SciML outputs directory.  I'm going to need someone to generate a deploy key and send it to me securely so I can encrypt it and add it to this pipeline configuration.

devmotion:
> Looks like the publishing step has some problems; we're trying to upload with an encrypted deploy key that is supposedly associated with https://github.com/TuringLang/TuringTutorialsOutput but I bet it's actually the encrypted deploy key for the SciML outputs directory. I'm going to need someone to generate a deploy key and send it to me securely so I can encrypt it and add it to this pipeline configuration.

I already assumed that it will fail and we have to add/modify the deploy key. What's the best way to send you the key? Or is it easier if I make you an admin of TuringTutorialsOutput temporarily?

staticfloat:
You can send it to me on slack, or make me an admin; either way is fine

PR #290, opened 2022-03-05
Update missing and broken tutorials

devmotion:
Currently, tutorial 4 is missing (build + deployment was canceled since additional commits were pushed to the master branch too quickly after the transition PR was merged) and tutorials 8, 10, 11, and 12 are broken due to outdated packages and code.

This PR fixes these issues but there are many more things (Turing features, deprecations, etc.) that should be updated that are not included in this PR.

When this PR is merged, all tutorials should run correctly, so even though some use outdated packages, suboptimal code etc. it should be possible to update turing.ml and switch to the TuringTutorialsOutput there.


yebai (approved):

devmotion:
The GP tutorial takes much longer to run than any other tutorial (~3.5 hours) even though some of the others in this PR are not the fastest ones either. I didn't check the details of the tutorials but it seems the GP could be improved quite a bit - I think the models could be made simpler and more efficient, and it seems in the sparse GP example the inducing points are not used but the model instead uses full covariance matrix of the observations. I didn't include any fixes for these issues in the PR as I wanted to focus on getting them up and running first.

rikhuijzer:
> The GP tutorial takes much longer to run than any other tutorial (~3.5 hours) even though some of the others in this PR are not the fastest ones either. I didn't check the details of the tutorials but it seems the GP could be improved quite a bit - I think the models could be made simpler and more efficient, and it seems in the sparse GP example the inducing points are not used but the model instead uses full covariance matrix of the observations. I didn't include any fixes for these issues in the PR as I wanted to focus on getting them up and running first.

Yes that one takes painfully long. Great work here David :+1: :+1: Seems that we're getting close to a really solid build system

PR #295, opened 2022-04-26
Run tutorials on Julia 1.7

rikhuijzer:
@devmotion Is this the place to change the Julia version to 1.7?

### DIFF
--- a/.buildkite/run_tutorial.yml
+++ b/.buildkite/run_tutorial.yml
@@ -18,7 +18,7 @@ steps:
             - BUILDKITE_S3_SECRET_ACCESS_KEY="U2FsdGVkX1+SPF81nkK7KQ64DsafSl0qq2iG7BsQs1xlTYEtZV3MqQl3l/NWaiocaEywZZFbAB5zpnKPD0xHTQ=="
             - BUILDKITE_S3_DEFAULT_REGION="U2FsdGVkX1/cORlxhXcxhja2JkqC0f8RmaGYxvGBbEg="
       - JuliaCI/julia#v1:
-          version: 1.6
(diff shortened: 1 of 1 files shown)


yebai (approved):

devmotion:
Yes, but there's one other line to update in this file.

However, I think it is a bit easier and requires fewer updates and rebuilds if we just use Julia LTS (i.e., keep using Julia 1.6). If the version is changed, we also have to update all Manifest files (IIRC 1.7 introduced a slightly different format) and most importantly have to update the cryptographic signature (it prevents any "random" PRs from changing the important parts of the buildkite setup, IIRC it can only be done by buildkite administrators).

devmotion:
> I fear that people then might conclude that Turing shouldn't be used on Julia 1.7 or that Julia 1.7 shouldn't be used even though the opposite is true.

What makes you think that users would assume this? I would have guessed most not even look at the system information but only if they can't rerun the tutorials. And otherwise just try to install Turing with whatever Julia version they use (and hence I think it might be more likely that they use a too old Julia version and hence Turing is hold back which in turn causes problems). We never state explicitly anywhere which Julia versions are supported (maybe we should and mention that LTS and all more recent Julia versions are supported?).

>  The 1.8 beta even shows 10% reduction in TTFX in many cases is what I've noticed during benchmarks

Since we run tutorials only once I don't think the hazzle of updating the packages and CI to new Julia versions is worth it. E.g., we also don't rerun the tutorials if newer non-breaking package versions are released.

rikhuijzer:
Whatever. I'm not here to have arguments. I leave the decision up to you

PR #296, opened 2022-04-26
Fix differential equation tutorial

devmotion:
I've wanted to do this for quite some time but unfortunately my computer broke while working on an update...

So here are some fixes for the differential equation tutorial.

Most notably,
- it fixes the `MvNormal` deprecations
- fixes the trajectories of posterior samples (the used the wrong parameters)
- it uses more "modern" and simpler SciML syntax
- specifies the solver for the DDE example correctly.

Preview: https://julialang-buildkite-artifacts.s3.amazonaws.com/turingtutorials/e3c8bc002f5223b50eb6e9c7902473c040159c0aa4d5942d979166fa05bcb422/html/10-bayesian-differential-equations/10_bayesian-differential-equations.html


devmotion:
@staticfloat I think there might be an issue with coppermind? https://buildkite.com/julialang/turingtutorials/builds/76#9b6b2925-254f-43b2-8401-9a7224be6376/205-207

devmotion:
Will be fixed by https://github.com/TuringLang/TuringTutorials/pull/298.

tutorials/10-bayesian-differential-equations/10_bayesian-differential-equations.jmd
```diff
@@ -13,312 +14,360 @@ using DifferentialEquations
...
     end
+
+    return nothing
 end
 
-model = fitlv(odedata, prob1)
+model = fitlv(odedata, prob)
 
-# This next command runs 3 independent chains without using multithreading.
+# Sample 3 independent chains with forward-mode automatic differentiation (the default).
```
yebai, line 109: Maybe add a link to the docs on how to change AD backends here?

yebai (approved):

tutorials/10-bayesian-differential-equations/10_bayesian-differential-equations.jmd
devmotion: It's shown how to change it (and change it back) below, is that sufficiently clear? Or should we add a link here? Or maybe rather below?

devmotion:
There's one remaining issue, namely that `pool_chain` is not exported by MCMCChains. It's a hack anyway, so I think I'll just go and fix https://github.com/TuringLang/MCMCChains.jl/issues/223. Then `pool_chain` is not needed anymore and the code becomes simpler.

tutorials/10-bayesian-differential-equations/10_bayesian-differential-equations.jmd
yebai: thanks for the clarification, that should be sufficient.

PR #302, opened 2022-04-29
Fix and simplify CompatHelper

devmotion:
The configuration was not correct but is fixed and simplified now. I tested it by running it manually, and it works (as the open PR shows).


yebai (approved):
thanks, @devmotion!

PR #303, opened 2022-05-02
Abort build process when error occurs

devmotion:
This should make it less likely to miss if a tutorial is broken. Without any options specified, Weave will just continue and include the error in the output. With this PR, Weave will throw an error instead and abort the build process.

Currently, the option is only specified for the hidden chunks with assertions but I think it's useful to set it for all chunks.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -36,3 +36,37 @@ TuringTutorials.weave(
     out_path_root=pkgdir,
     build=(:script, :html, :github, :notebook),
 )
+
+write(
+    joinpath(tutorials_dir, "test_error.jmd"),
+    """
+    ---
+    title: Test
+    author: TuringLang team
+    weave_options:
+      error : false
+    ---
+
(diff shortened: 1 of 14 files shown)


rikhuijzer:
Nope sorry I'm done with Weave. I've built this kind of logic in PlutoStaticHTML.jl to get rid of this kind of logic in the TuringTutorials repository. I'm not spending time to figure out **again** how to make CI fail on error.

devmotion:
> I'm not spending time to figure out **again** how to make CI fail on error.

Nothing to do for you here, it's an option in Weave and this PR enables it for all chunks :slightly_smiling_face: Just thought you might like it and approve the PR.

yebai (approved):
thanks, @devmotion - it looks good to me. 

I think both Weave and Pluto are great tools. Both might have some pros and cons. Maybe we can revisit this issue later -- we can compare the experiences between TuringTutorials and TuringGLM at some point.

PR #304, opened 2022-05-27
Use `filldist` in coinflip example

devmotion:
When updating some other tutorials, I noticed that maybe we should use (and hence somewhat promote) `filldist` already in the coinflip example.

I think it should be more well-known and used (even though I would still like to unify and extend it in Distributions), since it
- makes it easier to sample from the prior (without having to use `missing` and/or type parameters),
- should be efficient regardless of the AD backend (since it does not enforce the use of `for` loops but allows AD backends to use an optimized implementation)

### DIFF
--- a/tutorials/00-introduction/00_introduction.jmd
+++ b/tutorials/00-introduction/00_introduction.jmd
@@ -155,21 +155,23 @@ using MCMCChains
 First, we define the coin-flip model using Turing.
 
 ```julia
-@model function coinflip(y)
+# Unconditioned coinflip model with `N` observations.
+@model function coinflip(; N::Int)
     # Our prior belief about the probability of heads in a coin toss.
     p ~ Beta(1, 1)
 
-    # The number of observations.
-    N = length(y)
-    for n in 1:N
-        # Heads or tails of a coin are drawn from a Bernoulli distribution.
-        y[n] ~ Bernoulli(p)
-    end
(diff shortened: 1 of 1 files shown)


yebai (approved):
thanks, I'm happy with the changes. I am slightly unhappy with the fact we have multiple ways of defining a product distribution, e.g. using `arraydist` and `filldist`. I agree it is better to unify these heuristics (possibly also the dot notation) with the `Product` distribution in the longer term.

devmotion:
> e.g. using arraydist and filldist

We could deprecate `filldist` right away and instead use `arraydist(Fill(dist, N))` but it's a bit more (too?) verbose. Internally, actually `Distributions.product_distribution(Fill(dist, N))` will result in exactly the same objects and behaviour as `filldist(dist, N)` for almost all distributions: https://github.com/TuringLang/DistributionsAD.jl/blob/48c43f8e8062ba95542330735593b5275117e592/src/filldist.jl#L9-L12

yebai:
Personally, I'd depreciate `filldist` and encourage users to use `arraydist(Fill(dist, N))` from now. If it is too verbose, we can perhaps consider sugar syntax like `arraydist(dist, IID, N)`  where IID indicate the relationship between elements in the array?

devmotion:
IMO `arraydist(Fill(...))` would be a bit simpler and more consistent API. And also closer to something like `product` or `product_distribution` that we (hopefully) will use at some point.

PR #308, opened 2022-07-07
Adding link to condition/decondition docs in DynamicPPL

PavanChaggar:
Following up on a user comment in slack, I've added a link to the docs on condition/decondition syntax in the coin-flip example.

### DIFF
--- a/tutorials/00-introduction/00_introduction.jmd
+++ b/tutorials/00-introduction/00_introduction.jmd
@@ -173,7 +173,7 @@ coinflip(y::AbstractVector{<:Real}) = coinflip(; N=length(y)) | (; y);
 ```
 
 In the Turing model the prior distribution of the variable `p`, the probability of heads in a coin toss, and the distribution of the observations `y` are specified on the right-hand side of the `~` expressions.
-The `@model` macro modifies the body of the Julia function `coinflip` and, e.g., replaces the `~` statements with internal function calls that are used for sampling. 
+The `@model` macro modifies the body of the Julia function `coinflip` and, e.g., replaces the `~` statements with internal function calls that are used for sampling.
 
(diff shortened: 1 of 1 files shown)


tutorials/00-introduction/00_introduction.jmd
```diff
@@ -175,7 +175,8 @@ coinflip(y::AbstractVector{<:Real}) = coinflip(; N=length(y)) | (; y);
 In the Turing model the prior distribution of the variable `p`, the probability of heads in a coin toss, and the distribution of the observations `y` are specified on the right-hand side of the `~` expressions.
 The `@model` macro modifies the body of the Julia function `coinflip` and, e.g., replaces the `~` statements with internal function calls that are used for sampling.
 
-We combine the model with the observations:
+We have also defined a convenience function that makes use of the `condition` syntax (see ['Condition & decondition'](https://beta.turing.ml/DynamicPPL.jl/stable/api/#Condition-and-decondition) in DynamicPPL.jl
```
devmotion, line 178: I guess the title is not particularly helpful here. Maybe we could just refer to
```suggestion
We have also defined a convenience function that makes use of the `condition` syntax (see the [documentation of DynamicPPL.jl](https://beta.turing.ml/DynamicPPL.jl/stable/api/#Condition-and-decondition)
```

devmotion: Another general comment is that maybe it would be better to move the code to the next code block so that we can explain the conditioning before it is used.

devmotion: I applied the suggestions, what do you think @PavanChaggar?

yebai (approved):

yebai:
many thanks @PavanChaggar and @devmotion !

PR #316, opened 2022-07-19
Don't choose a sensealg

ChrisRackauckas:
I don't trust users to choose it right.


yebai (approved):

devmotion:
I just skimmed through the tutorial and am confused now -  isn't the main purpose of this section to demonstrate how to use `sensealg`? All other sections already use the default algorithm, so the explanations in this section seem a bit confusing now.

ChrisRackauckas:
All of the other sections use forward mode. It should just describe reverse mode, but it should let to the sensealg choice section, but probably not hardcode choose one.

devmotion:
> All of the other sections use forward mode.

Ah that's what I missed when skimming through the tutorial 👍

PR #318, opened 2022-08-10
Add MiniTuring tutorial.

sunxd3:


sunxd3:
This tutorial was written as a personal practice as I tried to understand DynamicPPL. As suggested by @yebai, I made this PR, in case someone thinks it is valuable for future Turing contributors. 

The code was verified in a local jupyter notebook setting. I tried to `weave` the jmd file locally but can't get over the `UndefVarError`. The stack trace is long, and I don't know if it is helpful or necessary to post the issue. Possible fixes are welcome. 

As suggested by @phipsgabler, an example with `macroexpand` and further explanation may be desirable, but I failed to find a good example for this purpose that is concise and doable with the limited functionality supported by the code. Suggestions on this is very welcomed too.

tutorials/14-minituring/14_minituring.jmd
```diff
@@ -0,0 +1,300 @@
...
+
+function _assume(__context__::MHTransitionContext, __varinfo__, sampler, dist, var_id)
+    old_value = __varinfo__.values[var_id]
+    # propose a random-walk step, i.e, add the current value to a random value sampled from
+    # a Normal distribution centered at 0
+    new_value = old_value + rand(__context__.rng, sampler.proposal)
+    log_p = Distributions.loglikelihood(dist, new_value)
+    update_varinfo!(__varinfo__, var_id, new_value, log_p)
+    return new_value
+end
```
devmotion, line 200: A bit simpler and with fewer underscores:
```suggestion
function assume(context::SamplingContext, varinfo, sampler::MHSampler, dist, var_id)
    old_value = varinfo.values[var_id]
    # propose a random-walk step, i.e, add the current value to a random value sampled from
    # a Normal distribution centered at 0
    value = old_value + rand(context.rng, sampler.proposal)
    logp = Distributions.loglikelihood(dist, new_value)
    update_varinfo!(varinfo, var_id, value, logp)
    return value
end
```

tutorials/14-minituring/14_minituring.jmd
```diff
@@ -0,0 +1,300 @@
...
+# Defining functions to compute values to determine if accept the proposal
+function logα(old_vi, new_vi, args, proposal)
+    return compute_joint(new_vi) - compute_joint(old_vi) 
+    + compute_condition_log(old_vi, new_vi, args, proposal) - 
+    compute_condition_log(new_vi, old_vi, args, proposal)
+end
+    
+function compute_condition_logprob(vi, condition_vi, args, proposal)
+    return reduce(+, [Distributions.loglikelihood(proposal, value1 - value2) 
+            for ((key, value1), (_, value2)) in zip(vi.values, condition_vi.values) if key ∉ args])
```
devmotion, line 228: This is a bit complicated. It also seems possibly wrong since it is not guaranteed that the keys of `value1` and `value2` are the same since `Dict` is not an ordered dictionary.
An initial simplification would be e.g.
```suggestion
    values1 = vi.values
    values2 = condition_vi.values
    return sum([loglikelihood(proposal, values1[key] - value2[key]) for key in keys(values1) if !(key in args)])
```

tutorials/14-minituring/14_minituring.jmd
```diff
@@ -0,0 +1,300 @@
...
+end
+
+function compute_joint(vi)
+    return reduce(+, Base.values(vi.log_ps))
+end
```
devmotion, line 233:
```suggestion
compute_joint(vi) = sum(values(vi.logps))
```

tutorials/14-minituring/14_minituring.jmd
```diff
@@ -0,0 +1,300 @@
...
+    ctx = SamplingContext()
+    model.f(model, vi, ctx, model.data...)
+    return Transition(vi), vi
+end
+
+# Defining functions to compute values to determine if accept the proposal
+function logα(old_vi, new_vi, args, proposal)
+    return compute_joint(new_vi) - compute_joint(old_vi) 
+    + compute_condition_log(old_vi, new_vi, args, proposal) - 
+    compute_condition_log(new_vi, old_vi, args, proposal)
```
devmotion, line 223: The `compute_` prefixes are quite verbose and not very idiomatic (I'd say), so maybe just use `logjoint` and `condition_logprob` or `transition_logprob`?

tutorials/14-minituring/14_minituring.jmd
```diff
@@ -0,0 +1,300 @@
...
+function compute_joint(vi)
+    return reduce(+, Base.values(vi.log_ps))
+end
+
+# The following steps
+function AbstractMCMC.step(
+    rng::Random.AbstractRNG,
+    model::MiniModel,
+    sampler::MHSampler,
+    prev_state, # is just the old vi
```
devmotion, line 240: Even clearer with a type annotation?
```suggestion
    prev_state::VarInfo, # is just the old vi
```

tutorials/14-minituring/14_minituring.jmd
```diff
@@ -0,0 +1,300 @@
...
+    kargs...
+)
+    vi = prev_state
+    new_vi = deepcopy(vi)
+    ctx = MHTransitionContext()
+    model.f(model, new_vi, ctx, model.data...)
+    log_α = logα(vi, new_vi, model.args, sampler.proposal)
+    # @show(new_vi, log_α)
+    if log(rand(rng)) < min(0, log_α)
```
devmotion, line 249:
```suggestion
    if -randexp(rng) < log_α
```

tutorials/14-minituring/14_minituring.jmd
```diff
@@ -0,0 +1,300 @@
...
+    end
+end
+```
+
+Next, let's define a `Model` struct. In the actual implementation of the compiler in Turing, the model construction is done as the final step of the compilation. But for the sake of simplicity, we will construct the model manually. This also requires us to keep a field for the actually data in the `Model` struct. 
+
+```julia
+struct MiniModel<:AbstractMCMC.AbstractModel
+    f
+    sampler
```
devmotion, line 153: The `sampler` is not part of the model in Turing, and also not in the AbstractMCMC interface, but provided separately to `sample`. Remove it here?
```suggestion
```

tutorials/14-minituring/Manifest.toml
```diff
@@ -0,0 +1,874 @@
+# This file is machine-generated - editing it directly is not advised
+
+julia_version = "1.7.3"
```
devmotion, line 3: Can you regenerate the Manifest file with Julia 1.6?

tutorials/14-minituring/Project.toml
```diff
@@ -0,0 +1,7 @@
+[deps]
+AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
```
devmotion, line 2: Can you add a `[compat]` section with compat entries for the different packages as well? That makes updating and keeping track of breaking releases of these dependencies easier.

tutorials/14-minituring/14_minituring.jmd
```diff
@@ -0,0 +1,327 @@
...
+
+We will have the following model definition in our small language.
+
+```julia; eval=false
+@mini_model function m(x)
+    a ~ Normal(0.5, 1)
+    b ~ Normal(a, 2)
+    x ~ Normal(b, 0.5)
+end
+```
```
sunxd3, line 35: Bluestyle requests a return value, but a return value here doesn't make sense. Is there a way to bypass the reviewdog on this?

tutorials/14-minituring/14_minituring.jmd
```diff
@@ -0,0 +1,327 @@
...
+The name `context` comes from the "contextual dispatching" idea of [Cassette.jl](https://github.com/JuliaLabs/Cassette.jl).
+Every `context`s is defined to be a type in Julia, and the `observe` and `assume` functions will take a `AbstractContext` type as their first arguments.
+Then, at runtime, the Julia multiple-dispatch system will dispatch corresponding functions according to the type that is a specific `context`.
+
+**Note:** *Although the context system in this tutorial is inspired by Turing's, Turing's context system is battle-tested and much more complicated.
+Thus, readers are advised to refer to Turing's documentations and code if knowledge of Turing's context system is required.
+Also, it should be pointed out that the context system design in Turing is not in its definitive form as of the current version at the time of this notebook is written.*
+
+```julia
+abstract type AbstractContext end
```
sunxd3, line 85: I think keeping the `MHTransitionContext` has its benefits as it demonstrates the idea of using a `context` to modify the behavior of `assume` and `observe`.

tutorials/14-minituring/14_minituring.jmd
```diff
@@ -0,0 +1,327 @@
...
+
+function mini_model(expr)
+    def = MacroTools.splitdef(expr)
+
+    # `MacroTools.postwalk` is a utility function for traverse AST in post-order
+    def[:body] = MacroTools.postwalk(def[:body]) do sub_expr
+        # `MacroTools.@capture` provide a utility function for pattern matching
+        if MacroTools.@capture(sub_expr, var_ ~ dist_)
+            if var in def[:args]
+                return quote
```
sunxd3, line 132: Instead of `MacroTools.@q`, I just use `quote` now, it should be less confusing.

tutorials/14-minituring/14_minituring.jmd
```diff
@@ -0,0 +1,327 @@
...
+        end
+    end
+
+    def[:args] = vcat(
+        [
+            # Insert extra arguments
+            :(model::($MiniModel)),
+            :(varinfo::($VarInfo)),
+            :(context::($AbstractContext)),
+            :(sampler::($(AbstractMCMC.AbstractSampler))),
```
sunxd3, line 151: `sampler` is not a part of `MiniModel` now, still needed in `assume` functions for `MHTransitionContext`, so pass in as an argument.

tutorials/14-minituring/Project.toml
sunxd3: `[compat]` section added.

sunxd3 (commented):
@devmotion Sorry for taking so long to come back to this. Most of the problems should be fixed, can you give it another look?

tutorials/14-minituring/14_minituring.jmd
devmotion: We could add a line with `#! format: off` above and `#! format: on` below the code. This could be hidden, I assume, by putting it into a `julia; echo=false` code block. Alternatively (and probably easier) would be to add a line `return nothing` to the model function.

tutorials/14-minituring/14_minituring.jmd
devmotion: `SamplingContext` is already an example for that, isn't it? And you can still specialize `assume` and `observe` based on the sampler (which is how it is done in Turing anyway).

tutorials/14-minituring/14_minituring.jmd
```diff
@@ -0,0 +1,327 @@
...
+        ],
+        def[:args],
+    )
+
+    return Base.remove_linenums!(
+        quote
+            $(MacroTools.combinedef(def))
+        end,
+    )
```
devmotion, line 160: Maybe simply
```suggestion
    return MacroTools.combinedef(def)
```
? I don't think it's important to remove line numbers in the tutorial.

tutorials/14-minituring/14_minituring.jmd
```diff
@@ -0,0 +1,327 @@
...
+But for the sake of simplicity, we will construct the model manually.
+This also requires us to keep a field for the actually data in the `MiniModel` struct.
+
+```julia
+struct MiniModel{F} <: AbstractMCMC.AbstractModel
+    f::F
+
+    # Data
+    args
+    data # a tuple of all the data
```
devmotion, line 175:
```suggestion
struct MiniModel{F,A,D} <: AbstractMCMC.AbstractModel
    f::F
    args::A
    data::D # a tuple of all the data
```

tutorials/14-minituring/14_minituring.jmd
```diff
@@ -0,0 +1,327 @@
...
+In simple words, we need to implement a sub-type of `AbstractModel`, which we did in the code block above; a sub-type of `AbstractSampler`, which we'll do in the next code block; and `step` functions, following the sampler definition.
+
+```julia
+# Just a Normal distribution as proposal in this simple case
+struct MHSampler{P} <: AbstractMCMC.AbstractSampler where {P<:Normal}
+    proposal::P
+end
+
+struct Transition
+    varinfo::VarInfo
```
devmotion, line 194:
```suggestion
struct Transition{V}
    varinfo::V
```

tutorials/14-minituring/14_minituring.jmd
```diff
@@ -0,0 +1,327 @@
...
+
+> **Julia Sidebar**
+> We used the Julia package [AbstractMCMC.jl](https://github.com/TuringLang/AbstractMCMC.jl) here.
+> For the interface design, we refer the reader to the [documentation](https://beta.turing.ml/AbstractMCMC.jl/dev/).
+
+In simple words, we need to implement a sub-type of `AbstractModel`, which we did in the code block above; a sub-type of `AbstractSampler`, which we'll do in the next code block; and `step` functions, following the sampler definition.
+
+```julia
+# Just a Normal distribution as proposal in this simple case
+struct MHSampler{P} <: AbstractMCMC.AbstractSampler where {P<:Normal}
```
devmotion, line 189:
```suggestion
struct MHSampler{P<:Normal} <: AbstractMCMC.AbstractSampler
```

tutorials/14-minituring/14_minituring.jmd
```diff
@@ -0,0 +1,327 @@
...
+**Note:** *Although the context system in this tutorial is inspired by Turing's, Turing's context system is battle-tested and much more complicated.
+Thus, readers are advised to refer to Turing's documentations and code if knowledge of Turing's context system is required.
+Also, it should be pointed out that the context system design in Turing is not in its definitive form as of the current version at the time of this notebook is written.*
+
+```julia
+abstract type AbstractContext end
+```
+
+```julia
+struct SamplingContext{R} <: AbstractContext where {R<:Random.AbstractRNG}
```
devmotion, line 89:
```suggestion
struct SamplingContext{R<:Random.AbstractRNG} <: AbstractContext
```

tutorials/14-minituring/14_minituring.jmd
```diff
@@ -0,0 +1,327 @@
...
+struct Transition
+    varinfo::VarInfo
+end
+```
+
+We hard-coded the proposal step as part of a `context` implementation.
+In Turing, to accommodate more use-cases, the `proposal` function is abstracted out.
+
+```julia
+struct MHTransitionContext{R} <: AbstractContext where {R<:Random.AbstractRNG}
```
devmotion, line 202:
```suggestion
struct MHTransitionContext{R<:Random.AbstractRNG} <: AbstractContext
```

tutorials/14-minituring/14_minituring.jmd
```diff
@@ -0,0 +1,327 @@
...
+    model.f(model, vi, ctx, sampler, model.data...)
+    return Transition(vi), vi
+end
+
+# Defining functions to compute values to determine if accept the proposal
+function condition_logprob(vi, condition_vi, args, proposal)
+    return sum([
+        Distributions.loglikelihood(proposal, vi.values[key] - condition_vi.values[key]) for
+        key in keys(vi.values) if !(key in args)
+    ])
```
devmotion, line 243: One can avoid allocating an array here:
```suggestion
    return sum(
        Distributions.loglikelihood(proposal, vi.values[key] - condition_vi.values[key]) for
        key in keys(vi.values) if !(key in args)
    )
```

tutorials/14-minituring/14_minituring.jmd
```diff
@@ -0,0 +1,327 @@
...
+    sampler::MHSampler,
+    prev_state::VarInfo, # is just the old vi
+    kargs...,
+)
+    vi = prev_state
+    new_vi = deepcopy(vi)
+    ctx = MHTransitionContext()
+    model.f(model, new_vi, ctx, sampler, model.data...)
+    log_α = logα(vi, new_vi, model.args, sampler.proposal)
+    # @show(new_vi, log_α)
```
devmotion, line 267:
```suggestion
```

tutorials/14-minituring/14_minituring.jmd
devmotion: BTW I think it might be easier here to use `NamedTuple`s as `data` instead of tuples and remove the separate `args` field:
```suggestion
struct MiniModel{F,D} <: AbstractMCMC.AbstractModel
    f::F
    data::D # a NamedTuple of all the data
```

tutorials/14-minituring/14_minituring.jmd
sunxd3: `julia; echo=false` will hide the code in the output; so just `return nothing` for now.

tutorials/14-minituring/14_minituring.jmd
```diff
@@ -0,0 +1,314 @@
...
+Then, at runtime, the Julia multiple-dispatch system will dispatch corresponding functions according to the type that is a specific `context`.
+
+**Note:** *Although the context system in this tutorial is inspired by Turing's, Turing's context system is battle-tested and much more complicated.
+Thus, readers are advised to refer to Turing's documentations and code if knowledge of Turing's context system is required.
+Also, it should be pointed out that the context system design in Turing is not in its definitive form as of the current version at the time of this notebook is written.*
+
+```julia
+struct SamplingContext{R<:Random.AbstractRNG,S<:AbstractMCMC.AbstractSampler}
+    rng::R
+    sampler::S
```
sunxd3, line 88: `sampler` is a part of `context` now, which is what DPPL does.

sunxd3 (commented):
@devmotion fixed, give it another review?

tutorials/14-minituring/14_minituring.jmd
```diff
@@ -0,0 +1,314 @@
...
+```
+
+Given data `x = 3.0`.
+
+```julia
+# Manually defining the model.
+model = MiniModel(m, (x=3.0,))
+
+# `sample` function is implemented as part of `AbstractMCMC.jl`
+samples = sample(model, MHSampler(Normal(0, 1)), 1000000)
```
sunxd3, line 294:
```suggestion
sample(model, MHSampler(Normal(0, 1)), 1000000)
```

phipsgabler (approved):

yebai:
thanks, @sunxd3!

PR #323, opened 2022-09-23
Improve Mini Turing tutorial

devmotion:
This PR simplifies the code a bit and fixes some typos and IMO slightly incorrect or misleading statements.

PR #324, opened 2022-09-24
Fix Turing MH sampler

devmotion:
I noticed that I had used the wrong syntax for the MH sampler in Turing in #323. I wanted to use the same random walk proposal as for the mini turing example but currently it uses static `Normal(0, 1)` proposals. This explains also the low ESS.

It seemed easiest (and probably it's one of the most efficient approaches) to just specify the covariance matrix of the random walk proposals, and it should be more efficient to specify explicitly that it is an identity matrix. Alternatively, but IMO less concise and/or efficient, one could use `MH(:a => x -> Normal(x, 1), :b => x -> Normal(x, 1))` or `MH(:a => AdvancedMH.RandomWalkProposal(Normal()), :b => AdvancedMH.RandomWalkProposal(Normal()))`.


devmotion:
Now the ESS estimates are similar: https://julialang-buildkite-artifacts.s3.amazonaws.com/turingtutorials/6f982592a7399e505fe00b53509b4a75a0d920f13033c375dbda8539d62d2299/html/14-minituring/14_minituring.html

yebai (approved):

yebai:
> It seemed easiest (and probably it's one of the most efficient approaches) to just specify the covariance matrix of the random walk proposals

That is true for this simple example, which is indeed a two-dimensional Gaussian. However, I don't think we need to be optimal with the sampler here -- the main purpose is to illustrate how the DSL works. 

Many thanks for the quick fixes and sorry that I didn't catch it in #323.

PR #326, opened 2022-10-17
Bayesian Time Series Tutorial

enweg:
Just changed the expression for a multiplicative decomposition to use a product instead of a sum.


devmotion (approved):
Thanks!

PR #327, opened 2022-10-18
New Documentation Structure

cgbotta:
As discussed with @yebai, we are porting all the `md` files in the "Using Turing" section of the public-facing site to `jmd`. This entails porting everything from [https://github.com/TuringLang/Turing.jl/tree/master/docs/src/using-turing](here) to `TuringTutorials/docs`. We are going to mimic the structure of the `TuringTutorials/tutorials` directory.

I began this process by adding a `source_folder` argument to the weave functions so that they can be used for weaving both tutorials and the new documentation (docs) folder. I also ported the first lesson. 

Open questions I have are:
- I have not opened any PRs for the Turing.jl repo - what changes need to be made on that end to recognize and render the new documentation structure on the site?
- I will also update the README instructions for weaving once this pattern is finalized


yebai:
@devmotion how do we add a new folder (e.g. `TuringTutorials/docs/`) to the build workflow?

docs/00-getting-started/00_getting-started.jmd
```diff
@@ -0,0 +1,78 @@
+---
+title: Getting Started
+permalink: /:collection/:name/
```
yebai, line 3: Would these fields of `permalink` and `redirect_from` still work with `jmd`?

docs/00-getting-started/00_getting-started.jmd
```diff
@@ -0,0 +1,78 @@
...
+### Install Turing.jl
+
+Turing is an officially registered Julia package, so you can install a stable version of Turing by running the following in the Julia REPL:
+
+```julia
+using Pkg
+Pkg.add("Turing")
+```
+
+You can check if all tests pass by running
```
yebai, line 28:
```suggestion
You can check if all tests pass by running (it might take a long time)
```

docs/00-getting-started/00_getting-started.jmd
```diff
@@ -0,0 +1,78 @@
...
+```julia
+@model function gdemo(x, y)
+    s² ~ InverseGamma(2, 3)
+    m ~ Normal(0, sqrt(s²))
+    x ~ Normal(m, sqrt(s²))
+    return y ~ Normal(m, sqrt(s²))
+end
+```
+
+Then we can run sampler a sampler to collect results. In this case it is a Hamiltonian Monte Carlo sampler
```
yebai, line 56:
```suggestion
Then we can run a sampler to collect results. In this case, it is a Hamiltonian Monte Carlo sampler
```

src/TuringTutorials.jl
```diff
@@ -91,23 +92,23 @@ function weave(
 end
 
 # Weave all tutorials
```
yebai, line 94:
```suggestion
# Weave all tutorials and documentations
```

src/TuringTutorials.jl
```diff
@@ -91,23 +92,23 @@ function weave(
...
-function weave(; kwargs...)
-    for folder in readdir(joinpath(REPO_DIR, "tutorials"))
-        weave(folder; kwargs...)
+function weave(source_folder::AbstractString; kwargs...)
```
yebai, line 95: I think we still need a default no-argument `weave` function that processes all docs and tutorials.

docs/00-getting-started/00_getting-started.jmd
cgbotta: I copied this structure from all the files in `TuringTutorials/tutorials`, but I am assuming this is read by the build process. Perhaps @devmotion could also advise here?

devmotion:
> @devmotion how do we add a new folder (e.g. TuringTutorials/docs/) to the build workflow?

To the buildkite CI? One has to change the configuration in https://github.com/TuringLang/TuringTutorials/tree/master/.buildkite, the structure in https://github.com/TuringLang/TuringTutorialsOutput/ (everything has to be moved to subfolders `/docs/` and `/tutorials/` I assume), and therefore also the integration in https://github.com/TuringLang/turing.ml.

I wonder if it would be simpler, and docs and tutorials more cleanly separated, if one would use a separate repo instead of multiple subfolders. Or if one should just move everything into `/docs` (or `/tutorials`, whatever seems more reasonable).

src/TuringTutorials.jl
cgbotta: Done

yebai:
>  Or if one should just move everything into /docs (or /tutorials, whatever seems more reasonable)

I also think it is simpler to setup and maintain to have both docs and tutorials under one new folder (e.g. `docs`). We can distinguish between them by setting up the [navigation](https://github.com/TuringLang/turing.ml/blob/master/_data/toc.yml) bar.

cgbotta:
Thank you both for the input - I agree with one main `docs` folder

Is there a preference for me to port all of the `md` docs to `jmd` in this PR, or a separate PR for each one?

yebai:
> Is there a preference for me to port all of the md docs to jmd in this PR, or a separate PR for each one?

Let's setup the workflow first. We can have a few more additional PRs to transfer all other doc md files.

cgbotta:
After looking through the build process, there is a fair bit of refactoring involved across many files to change the name of the folder from `tutorials` to `docs`. Of course, I am more than happy to do that if we decide it is best. However, I wanted to discuss the changes I made in the latest commit because it seems like a simpler solution.

By leaving the folder named `tutorials`, I could just append `"tutorial-"` to the existing files, then the documentation files can begin with `"docs-"`. 

All that should be needed after this is to update the navigation bar. We could test it by only changing the first item https://turing.ml/v0.21/docs/using-turing/get-started

devmotion:
If we keep the `/tutorials` folder, couldn't we just not rename the existing tutorials? `tutorials/tutorials-00-...` seems a bit unnecessarily complicated.

cgbotta:
Good point - I am all for simplicity here 👍  Perhaps `tutorials/docs-00-...` is still worthwhile, but I agree that writing tutorials again is not necessary. Or if it really doesn't matter, I will just keep incrementing and start with `tutorials/15-getting-started` and not worry about distinguishing between old and new anywhere other than the nav bar

yebai:
> Perhaps `tutorials/docs-00-...` is still worthwhile

That sounds like a good tradeoff between minimal disruptions and being descriptive for now. Let us be incremental here, we can rename the tutorials to reflect a new naming scheme when things settle down a bit.

devmotion:
Buildkite reports an error:
```
Unable to find project for tutorials/docs-00-getting-started/00_getting-started.jmd
```
Could you add Project.toml and Manifest.toml (with Julia 1.6) and add the folder to the CompatHelper Github action?

tutorials/docs-00-getting-started/00_getting-started.jmd
```diff
@@ -0,0 +1,78 @@
...
+Turing is an officially registered Julia package, so you can install a stable version of Turing by running the following in the Julia REPL:
+
+```julia
+using Pkg
+Pkg.add("Turing")
+```
+
+You can check if all tests pass by running (it might take a long time)
+
+```julia
```
yebai, line 30:
```suggestion
```{julia; eval=false}
```

yebai: It might be better to disable running the entire Turing test suite while compiling this tutorial...

tutorials/docs-00-getting-started/00_getting-started.jmd
```diff
@@ -0,0 +1,78 @@
...
+chn = sample(gdemo(1.5, 2), HMC(0.1, 5), 1000)
+```
+
+We can use `describe` to summarise results
+
+```julia
+describe(chn)
+```
```
devmotion, line 66: Better to remove this since the desired summary statistics are shown as output to the sample call above but not when calling `describe` here:

```suggestion
```

Actually I fixed this in quite a few tutorials recently.

yebai:
@devmotion Do you know why artefacts are not generated?

devmotion:
They are generated, I just looked at them.

tutorials/docs-00-getting-started/00_getting-started.jmd
devmotion: Maybe another possible improvement might be to rather compare the results and plots with the exact posterior distribution instead of showing `savefig` which is more a Plots.jl detail.

tutorials/docs-00-getting-started/00_getting-started.jmd
```diff
@@ -0,0 +1,78 @@
+---
+title: Getting Started
+permalink: /:collection/:name/
```
yebai, line 3:
```suggestion
permalink: docs/using-turing/get-started
```

This will keep the URL unchanged.

tutorials/docs-00-getting-started/00_getting-started.jmd
yebai: It sounds like an interesting exercise for Bayesian inference 101. Can you work this out? @cgbotta

yebai (approved):

tutorials/docs-00-getting-started/00_getting-started.jmd
cgbotta: I think that is a great idea. I changed a decent bit of the tutorial so that the format is a bit easier to follow for beginners (in my opinion) and has the user generate data so that they can observe the model learning the correct parameters. 

Let me know your thoughts!

cgbotta: Ah I misinterpreted - apologies. I can move forward with this approach 👍

cgbotta: Okay this has been a good exercise, but I do have a question since I am still learning. 

If I understand correctly, having an unknown mean and variance results in having a prior with 4 parameters to update using the below equations. 

image

That all makes sense, though I am having trouble finding a suitable prior distribution with which to update. `NormalInverseGamma` does not seem to be available, but I do see `Distributions.NormalInverseGaussian`. Would that be the correct prior in this case?

devmotion: There's no normal-inverse gamma distribution in Distributions IIRC. However, I don't think you need it either: In the model it's factorized into the distribution of sigma^2 (first line of the model) and distribution of mu conditioned on sigma^2 (second line) (you can compare it with https://en.wikipedia.org/wiki/Normal-inverse-gamma_distribution#Definition, in our example we use N-IG(mu=0, lambda=1, alpha=2, beta=3)). For checking the MCMC approximations of the posterior, for a quick check I guess the mean of of the updated N-IG distribution might be sufficient (given e.g. in the summary box on Wikipedia), that's also what we usually check in the Turing.jl tests. Alternatively, you could use that the marginal distributions are simply an IG distribution (for the variance) and a student t distribution (for the mean): https://en.wikipedia.org/wiki/Normal-inverse-gamma_distribution#Marginal_distributions

cgbotta: Thanks for the helpful pointers - PR is updated

yebai:
@cgbotta it seems the build is broken. https://buildkite.com/julialang/turingtutorials/builds/185#018414e3-f154-4902-a5b3-f01bacda6dd8

cgbotta:
> @cgbotta it seems the build is broken. https://buildkite.com/julialang/turingtutorials/builds/185#018414e3-f154-4902-a5b3-f01bacda6dd8

I was just having a look at that. I see this:

```
/etc/buildkite-agent/plugins/github-com-staticfloat-coppermind-buildkite-plugin/lib/common.sh:
 line 11: /etc/buildkite-agent/plugins/github-com-staticfloat-coppermind-buildkite-plugin/lib/bash-tools/lib/buildkite.sh: No such file or directory
```

Not quite sure what that is implying is wrong, but I am digging through some of the build kite pipelines. Could it be an issue with the commit a week ago that added coppermind v1?

yebai:
I think that's correct. Updating this PR with changes from the `master` branch seems to fix it.

yebai (approved):

tutorials/docs-00-getting-started/00_getting-started.jmd
```diff
@@ -0,0 +1,84 @@
...
+In this case, because we use the [normal-inverse gamma distribution](https://en.wikipedia.org/wiki/Normal-inverse-gamma_distribution)
+as a [conjugate prior](https://en.wikipedia.org/wiki/Conjugate_prior), we can compute
+its updated mean as follows:
+
+```julia
+s² = InverseGamma(2, 3)
+m = Normal(0, 1)
+data = [1.5, 2]
+
+true_posterior_mean = (m.σ * m.μ + length(data) * mean(data)) / (m.σ + length(data))
```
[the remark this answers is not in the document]
cgbotta: @yebai Pushed some updates to show the posterior of `s²` as well

yebai:
@cgbotta Thanks for the updates. Let's replace the moment-matched `Normal` approximation to the posterior with a KDE estimation -- see the [`density` ](https://beta.turing.ml/MCMCChains.jl/stable/statsplots/#Density) function from `MCMCChains`.

tutorials/docs-00-getting-started/00_getting-started.jmd
```diff
@@ -0,0 +1,102 @@
+---
+title: Getting Started
+permalink: /docs/using-turing/get-started
```
yebai, line 3: Will this `permalink` work when we use version control? If not, is there a way to insert the `version` (e.g. https://turing.ml/v0.21/tutorials/00-introduction/) to permalinks in Jekyll?

Summon our god of Jekyll @cpfiffer

cpfiffer: Yes, I think it should go through just fine because the root includes the version number.

cgbotta:
>

cgbotta:
> @cgbotta Thanks for the updates. Let's replace the moment-matched `Normal` approximation to the posterior with a KDE estimation -- see the [`density` ](https://beta.turing.ml/MCMCChains.jl/stable/statsplots/#Density) function from `MCMCChains`.

Thanks Hong - forgive me if this is obvious, but that be show anything different from the posterior densities given when running `plot(chn)` a bit earlier in the tutorial?

yebai:
> Thanks Hong - forgive me if this is obvious, but that be show anything different from the posterior densities given when running plot(chn) a bit earlier in the tutorial?

I think `plot` will show multiple things, including the `density` plot. So, yes, the plot function is the same. But, notice that the samples from the analytical posterior are exact -- we know how to sample form a Normal-Inverse-Gamma distribution in closed form. So we are using the KDE `desity` plot to compare the exact samples against MCMC samples. If the MCMC samples converge well, they should match the exact samples. Does that answer your question? If not, we can chat a bit more when we meet.

tutorials/docs-00-getting-started/00_getting-started.jmd
```diff
@@ -0,0 +1,110 @@
...
+    samples = []
+    for i in 1:iterations
+        sample_variance = rand(InverseGamma(alpha, beta), 1)
+        sample_x = rand(Normal(mean, sqrt(sample_variance[1]) / lambda), 1)
+        sanples = append!(samples, sample_x)
+    end
+    return samples
+end
+
+samples = sample_posterior(updated_alpha, updated_beta, mean_exp, 2, 1000);
```
yebai, line 104:
```suggestion
analytical_samples = sample_posterior(updated_alpha, updated_beta, mean_exp, 2, 1000);
```

yebai:
Thanks, @cgbotta!

PR #328, opened 2022-10-19
.buildkite: run_tutorial: use `v1` of `coppermind` plugin

thazhemadam:
Since `v1` release of the `coppermind` plugin has been made, explicitly specify that the `v1` of the plugin must be used.


devmotion:
@staticfloat could you please update the keys? 🙂

devmotion (approved):

PR #349, opened 2022-11-24
Yongchao/update variational inference tutorial

YongchaoHuang:


tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -301,19 +301,25 @@ end
...
 end
 ```
 
 ```julia
 # Remove the model column.
 select!(data, Not(:Model))
 
-# Standardize our dataset.
-(std_data, data_arr) = standardize(Matrix(data))
+# NB: we need to split the data into train and test sets, then do standardization/normalisation. If you swap the sequence, there will be an info leakage, which means information from test sets is leaked into the trianing and leads to over-optimistic about the model performance but worse at generalisation.
```
yebai, line 315: Good to know.

yebai:
@YongchaoHuang can you take a look at why CI (buildkite, see above) is broken for this PR, please?

YongchaoHuang:
> @YongchaoHuang can you take a look at why CI (buildkite, see above) is broken for this PR, please?

sure, I will have an investigation later.

YongchaoHuang:
> @YongchaoHuang can you take a look at why CI (buildkite, see above) is broken for this PR, please?

Got it: I modified the unstandardize() function - the std and mean of an DataFrame raise the error. It's fixed by converting `Matrix(DataFrame)` or `Array(DataFrame)` .

yebai:
@YongchaoHuang I fixed all the formatting issues by accepting all the suggested changes by `JuliaFormatter`.

YongchaoHuang:
@torfjelde @yebai your review on these changes would be very much appreciated.

yebai:
@YongchaoHuang `buildkite` CI is not passing, can you take a look please?

YongchaoHuang:
> @YongchaoHuang `buildkite` CI is not passing, can you take a look please?

Yes, thanks Hong. I am looking into it.

YongchaoHuang:
> > @YongchaoHuang `buildkite` CI is not passing, can you take a look please?
> 
> Yes, thanks Hong. I am looking into it.

The Buildkite error says 'ERROR: LoadError: UndefVarError: histogram not defined'. 
`histogram` plot is from `Plots`, which is included already. It worked on my local machine.
I just replaced `histogram` by `Plots.histogram`. 
Also included `using Distributions`. Let's see how it works.

YongchaoHuang:
> > @YongchaoHuang `buildkite` CI is not passing, can you take a look please?
> 
> Yes, thanks Hong. I am looking into it.

@yebai passed now.

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -119,7 +119,8 @@ q isa MultivariateDistribution
 This means that we can call `rand` to sample from the variational posterior `q`
 
 ```julia
-rand(q)
+using Plots
+Plots.histogram(rand(q, 1_000)[1, :])
```
torfjelde, line 123: Do you know why this is `Plots.` specification is necessary? We should probably figure out why this is needed before we include it, since Plots should be exporting this method.

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -349,6 +362,25 @@ test = Matrix(test_cut[:, remove_names]);
 end;
 ```
 
+```julia
```
torfjelde, line 365: Why are we defining two models here?

As far as I can tell, the previous definition of `linear_regression` is not being used?

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -435,6 +471,7 @@ avg[union(sym2range[:coefficients]...)]
 With a bit of work (this will be much easier in the future), we can also visualize the approximate marginals of the different variables, similar to `plot(chain)`:
 
 ```julia
+using Plots, StatsPlots
```
torfjelde, line 474: A couple of things here:
- `StatsPlots` is meant to be a drop-in for `Plots`, i.e. we shouldn't have to import _both_ but only importing `StatsPlots` should suffice.
- It shouldn't be necessary to import `StatsPlots` (or `Plots`) in every block we want to use their functionality. Once should be enough.

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -620,15 +672,15 @@ scatter(
...
 xaxis!(1:size(test, 1))
-ylims!(95, 140)
+ylims!(10, 40)
 title!("MCMC (NUTS)")
 ```
 
 Indeed we see that the MCMC approach generally provides better uncertainty estimates than the mean-field ADVI approach! Good. So all the work we've done to make MCMC fast isn't for nothing.
 
-## Alternative: provide parameter-to-distribution instead of `q` with`update` implemented
+## Alternative: provide parameter-to-distribution instead of $q$ with $update$ implemented
```
torfjelde, line 683:
```suggestion
## Alternative: provide parameter-to-distribution instead of $q$ with `update` implemented
```

torfjelde (changes requested):
Looks good overall! But I would like to resolve the namespace issue before we merge. There are also a couple of redundant changes that can be reverted.

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: yes, it doesn't seem to be necessary to explicitly include the `Plots.` prefix; however, in a previous commit, buildkite failed with `ERROR: LoadError: UndefVarError: Plots not defined` (see [here](https://buildkite.com/julialang/turingtutorials/builds/327#01850879-f563-4a6f-8f47-3de0f60ccb3e)). So I explicitly added `Using Plots` and `Plots.histogram` in each use. 
I am investigating why this is needed.

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -296,28 +299,38 @@ end
...
 end
 
 # Another helper function to unstandardize our datasets.
 function unstandardize(x, orig)
-    return (x .+ mean(orig; dims=1)) .* std(orig; dims=1)
+    return x .* std(Array(orig); dims=1) .+ mean(Array(orig); dims=1)
+end
+
+function unstandardize(x, mean_train, std_train)
+    return x .* std_train .+ mean_train
```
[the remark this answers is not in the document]
YongchaoHuang: Yes, in this particular context, `orig` is a `Matrix` type. I added the Array coercion just in case - too safety aware like writing a Python def to reduce test efforts.
I shall remove the `Array` coercion then, as it's not written as a generic function but for this particular use.

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: In the old version, the `linear_regression` function has signature `(x, y, n_obs, n_vars)`, but `n_obs` is not used because of the use of `y ~ MvNormal(mu, σ² * I)`; in the newly defined `linear_regression(x, y, n_obs, n_vars)`, it iterates through `n_obs` and uses `y[i] ~ Normal(mu[i], sqrt(σ²))` (though could be less efficient). So I kept both versions for user to choose.
It seems not necessary to keep both - I just removed the first function.

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: > A couple of things here:

Yes totally agree with these points. 

> * `StatsPlots` is meant to be a drop-in for `Plots`, i.e. we shouldn't have to import _both_ but only importing `StatsPlots` should suffice.
I think in the original [version](https://github.com/TuringLang/TuringTutorials/blob/master/tutorials/09-variational-inference/09_variational-inference.jmd), both `Plots` and `StatsPlots` are imported, so I just leave as it is.
As pointed out, `StatsPlots` is a drop-in replacement for `Plots`, I just removed Plots to test it - will follow this up once buildkite is working.

> * It shouldn't be necessary to import `StatsPlots` (or `Plots`) in every block we want to use their functionality. Once should be enough.
Yes, this is again the same issue with the previous buildkite failure - Initially I just imported them once, it cannot find the `Plots` module without importing it in every julia code block. 
I am removing the redundant imports and see how it works again - will follow this up.

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -443,22 +480,29 @@ function plot_variational_marginals(z, sym2range)
...
+                titlefontsize=10,
+                label="",
+                primary=false,
+            )
             push!(ps, p)
         end
     end
 
-    return plot(ps...; layout=(length(ps), 1), size=(500, 1500))
+    return plot(ps...; layout=(length(ps), 1), size=(500, 1500), primary=false)
```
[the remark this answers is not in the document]
YongchaoHuang: I tried to disable the ylabel in the original plots, as they are not properly shown. 
In [this discourse](https://discourse.julialang.org/t/suppressing-the-label-when-using-plots-or-statsplots/27658), it says this could be achieved by setting `primary=false` as an argument in the `plot` function. 
It doesn't seem to do its [job](https://julialang-buildkite-artifacts.s3.amazonaws.com/turingtutorials/4aa906f01a50c95f91b989d5027856c68f7ba9e1e218c6fc08719bf3df2a3791/html/09-variational-inference/09_variational-inference.html) though.
Any suggestions to disable the ylabel (currently the word `density` is not properly displayed) in `plot()`?

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -598,16 +650,16 @@ scatter(
     size=(900, 500),
     markersize=8,
 )
-scatter!(1:size(test, 1), unstandardize(test_label, data.MPG); label="true")
+scatter!(1:size(test, 1), unstandardize(test_label, train_copy.MPG); label="true")
 xaxis!(1:size(test, 1))
-ylims!(95, 140)
+ylims!(10, 40)
```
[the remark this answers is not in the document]
YongchaoHuang: Yes, I think this is associated with the original version of `unstandardize()` function - it adds mean value and then multiplies std, which produced incorrectly large values.
This is corrected in the current version, and the scale of the predicted values now is smaller. 
This change also applies to any MSE calculations and discussions about comparison of the VI, MCMC, least squares results.

tutorials/09-variational-inference/09_variational-inference.jmd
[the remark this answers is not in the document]
torfjelde: > I tried to disable the ylabel in the original plots, as they are not properly shown.

Are you referring to the yticks being pushed out of the visible figure?

If so, those are the _ticks_, not the labels. `primary=false` has to do with the legend, not the `ylabel`. 

To fix how they're being displayed, I think you'll be better off messing around with the `margin` keyword argument, which is `1.0mm` by default. Try making this something like `2.0mm`.

> In this discourse, it says this could be achieved by setting primary=false as an argument in the plot function.

This is referring to the `label` used for the legend; not `ylabel`.

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -480,18 +501,23 @@ plot(chain)
...
     union(sym2range[:coefficients]...)...,
     union(sym2range[:intercept]...)...,
-    union(sym2range[:σ₂]...)...,
+    union(sym2range[:σ²]...)...,
 ]]
 ```
 
 ```julia
-mean(chain).nt.mean
+NUTS_mean = mean(chain).nt.mean
```
torfjelde, line 509:
```suggestion
mcmc_mean = mean(chain).nt.mean
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -480,18 +501,23 @@ plot(chain)
...
+```
+
+```julia
+plot(NUTS_mean; linestyle=:dot, label="NUTS")
```
torfjelde, line 513:
```suggestion
plot(mcmc_mean; linestyle=:dot, label="NUTS")
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -480,18 +501,23 @@ plot(chain)
...
+plot!(vi_mean; linestyle=:dot, label="VI")
 ```
 
 One thing we can look at is simply the squared error between the means:
 
 ```julia
-sum(abs2, mean(chain).nt.mean .- vi_mean)
+sum(abs2, NUTS_mean .- vi_mean)
```
torfjelde, line 520:
```suggestion
sum(abs2, mcmc_mean .- vi_mean)
```

tutorials/09-variational-inference/09_variational-inference.jmd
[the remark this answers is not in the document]
torfjelde: Gotcha :+1:

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -345,7 +354,9 @@ test = Matrix(test_cut[:, remove_names]);
 
     # Calculate all the mu terms.
     mu = intercept .+ x * coefficients
-    return y ~ MvNormal(mu, σ² * I)
+    for i in 1:n_obs
+        y[i] ~ Normal(mu[i], sqrt(σ²))
+    end
```
torfjelde, line 359: Is there a particular reason why this change was made? Why are we not using the vectorized implementation?

tutorials/09-variational-inference/09_variational-inference.jmd
torfjelde: >  Initially I just imported them once, it cannot find the Plots module without importing it in every julia code block.

Yeah so I get that this is the issue, but my point is that this seems to be a somewhat deeper issue with how we use buildkite or something indicating that each block is not evaluated in the same module? It seems like something we should fix rather than hack around, since otherwise we'll have to do this "import in every block" for every tutorial going forwards :confused:

tutorials/09-variational-inference/09_variational-inference.jmd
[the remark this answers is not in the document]
YongchaoHuang: Oh I see, so I should use `margin=2.0mm` instead. I am trying it out. Thanks for the brilliant thoughts.

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: great naming!

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: I think it might be useful to make explicit use of `n_obs` as it is set as an argument in the old version. Also may be more readable and insutitive for new commers (as many Stan/PyMC3 tutorials do).
Admittedly, vectorised version is more efficient - I just recerted to that.

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -296,28 +295,38 @@ end
...
 # Remove the model column.
 select!(data, Not(:Model))
 
-# Standardize our dataset.
-(std_data, data_arr) = standardize(Matrix(data))
-
 # Split our dataset 70%/30% into training/test sets.
-train, test = split_data(std_data, 0.7)
+train, test = split_data(data, 0.7)
+train_copy = copy(train) #  save a copy of the train data for later unstandardize use; alternatively, you can just save the statistics (mean and var) of the traiing data.
```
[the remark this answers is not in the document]
YongchaoHuang: Thanks for spotting this. Yes, in the original version the `train` and `test` are re-assigned later, as in Line 339 ad 340:
```
train = Matrix(train_cut[:, remove_names]);
test = Matrix(test_cut[:, remove_names]);
```
They were re-assigned as the standardized version of the original train and test sets. 
I personally would prefer using different names (i.e. create new variables) after each assignment - if it's not too memory intensive. I didn't make this change as the `train` and `test` variables are heavily used later, and re-factoring could be a bit work. 

So I save a copy of the unstandardized versions of `train` and `test` immediately after splitting, these copies are used later for transforming back the predicted values to the original space, as used as an argument of `unstandardize()`.

Again, it's also a safety-awareness thing, I sometimes keep this habit when writing data pipeline for DS projects, as if the important variables are overwritten, I have to re-start the workflow again.

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: Yes definitely need to look into this critical issue why the build environment doesn't make all imports store in the workspace - it seems each time it finishes executing a julia code block, it does some garbage collection thing? 
However, it just happens to the module `Plots`. Is it because it has got two choices, i.e. `Plots` and `StatsPlots`? and automatically clear up `Plots`? Maybe some dependency settings?

tutorials/09-variational-inference/09_variational-inference.jmd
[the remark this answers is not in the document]
torfjelde: I'm also not a fan of continual reassigment of a variable and very much agree that we should ideally not be doing this.

_But_ I don't think it's too helpful that we only do this _once_, and we give it a not-so-informative name like `train_copy`. If it's to separate between standardized and not standardized, I'd prefer to call it exactly that, i.e. `train_unstandardized`. Then it's utility becomes clear.

Btw, as a side-note, it's preferable if you let the original comment decide whether or not to "Resolve" a conversation (unless it's some trivial change). This makes it easier for me (the original commenter) when I come back to the PR to review your replies.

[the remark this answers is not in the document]
YongchaoHuang: Agreed, I just changed `train_copy` to `train_unstandardized`. This is much informative than the original one. 
Sorry I "resolved" the conversations without referencing you; I should leave the decision to the original commenter to decide from now on - a good learning.

tutorials/09-variational-inference/09_variational-inference.jmd
torfjelde: Sorry, just saw this comment. And yeah, it's probably because both `Plots` and `StatsPlots` are `export` ing the same functions, e.g. `histogram`, and so Julia requires you to use the namespace when calling the function. Hence if you just do `using StatsPlots` once and remove all `using Plots` statements, the issue should go away:)

YongchaoHuang: Thanks very much for the thoughts - yes I removed all `Plots` imports and only kept `StatsPlots`, it solves the problem.

tutorials/09-variational-inference/Project.toml
```diff
@@ -1,12 +1,14 @@
 [deps]
 Bijectors = "76274a88-744f-5084-9051-94815aaf08c4"
 ComponentArrays = "b0b7db55-cfe3-40fc-9ded-d10e2dbeff66"
+ConjugatePriors = "1624bea9-42b1-5fc1-afd3-e96f729c8d6c"
```
devmotion, line 4: I very strongly suggest not re-introducing ConjugatePriors. The package has not been updated in a while, does not support Julia 1.6 properly, and will cause major downgrades of Distributions and Turing. It's also not needed since all the analytical calculations can be performed in ~5 lines of code. Therefore I removed it in a PR last week.
```suggestion
```

YongchaoHuang: @devmotion Many thanks for letting me know this. I inherited it from the old version, and found some distribution methods are from it. I am looking into it.

YongchaoHuang: @devmotion Thanks for removing the `ConjugatePrior` and re-wrote the following lines:
```
p_μ = μₙ + sqrt(βₙ / (αₙ * κₙ)) * TDist(2 * αₙ)
p_μ_pdf = z -> pdf(p_μ, z)
```
I guess in the first line you want to do the linear transform `y=a+b*x` (correct me if I am wrong), which induces the current building failure `No method matching *(float64, Distribution)` - for some reason it cannot build a linear transform function `p_μ(x)` by scaling a distribution (even though we take the sampling view). I therefore tried re-writing it in the following form, which avoids the building error.
```
p_μ0 = TDist(2 * αₙ)
p_μ_pdf = z -> pdf(p_μ0, (z-μₙ)/sqrt(βₙ / (αₙ * κₙ)))/sqrt(βₙ / (αₙ * κₙ))
```
This is constructed following the principle of probability of transformed variables:
f_Y(y) = 1/b * f_X((y-a)/b)
Not sure if my understanding is correct; it would be helpful if you can confirm it. Many thanks.

devmotion: Just revert your latest changes. The current code works just fine *if* you use a somewhat recent version of Distributions. As I said, re-introducing ConjugatePriors downgrades Distributions *and* Turing a lot.

So my suggestion is:
- reverting the latest commit with your rewrites
- updating the Manifest.toml file by running `] up`

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -207,8 +204,8 @@ p_σ²_pdf = z -> pdf(p_σ², z)
 
 # marginal of μ
 # Eq. (91) in "Conjugate Bayesian analysis of the Gaussian distribution" by Murphy
-p_μ = μₙ + sqrt(βₙ / (αₙ * κₙ)) * TDist(2 * αₙ)
-p_μ_pdf = z -> pdf(p_μ, z)
+p_μ0 = TDist(2 * αₙ)
+p_μ_pdf = z -> pdf(p_μ0, (z - μₙ) / sqrt(βₙ / (αₙ * κₙ))) / sqrt(βₙ / (αₙ * κₙ))
```
devmotion, line 208: This change should not be needed:
```suggestion
p_μ = μₙ + sqrt(βₙ / (αₙ * κₙ)) * TDist(2 * αₙ)
p_μ_pdf = z -> pdf(p_μ, z)
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -19,19 +19,20 @@ q = vi(m, vi_alg)  # perform VI on `m` using the VI method `vi_alg`, which retur
 
 Thus it's no more work than standard MCMC sampling in Turing.
 
-To get a bit more into what we can do with `vi`, we'll first have a look at a simple example and then we'll reproduce the [tutorial on Bayesian linear regression](../../tutorials/5-linearregression) using VI instead of MCMC. Finally we'll look at some of the different parameters of `vi` and how you for example can use your own custom variational family.
+To get a bit more into what we can do with `vi`, we'll first have a look at a simple example and then we'll reproduce Turingthe [tutorial on Bayesian linear regression](../../tutorials/5-linearregression) using VI instead of MCMC. Finally we'll look at some of the different parameters of `vi` and how you for example can use your own custom variational family.
```
devmotion, line 22:
```suggestion
To get a bit more into what we can do with `vi`, we'll first have a look at a simple example and then we'll reproduce the [tutorial on Bayesian linear regression](../../tutorials/5-linearregression) using VI instead of MCMC. Finally we'll look at some of the different parameters of `vi` and how you for example can use your own custom variational family.
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -19,19 +19,20 @@ q = vi(m, vi_alg)  # perform VI on `m` using the VI method `vi_alg`, which retur
...
 using Random
 using Turing
 using Turing: Variational
+using StatsPlots, Measures
 
 Random.seed!(42);
 ```
 
-## Simple example: Normal-Gamma conjugate model
+# Simple example: Normal-Gamma conjugate model
```
devmotion, line 35: IIRC other tutorials use `##` as the highest-level heading:
```suggestion
## Simple example: Normal-Gamma conjugate model
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -232,14 +229,12 @@ xlims!(-0.25, 0.25)
 plot(p1, p2; layout=(2, 1), size=(900, 500))
 ```
 
-# Bayesian linear regression example using `ADVI`
+# Bayesian linear regression example using ADVI
```
devmotion, line 232:
```suggestion
## Bayesian linear regression example using ADVI
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -154,27 +155,23 @@ For that we need samples:
...
 ```julia
 p1 = histogram(samples[1, :]; bins=100, normed=true, alpha=0.2, color=:blue, label="")
-density!(samples[1, :]; label="s (ADVI)", color=:blue, linewidth=2)
-density!(samples_nuts, :s; label="s (NUTS)", color=:green, linewidth=2)
-vline!([var(x)]; label="s (data)", color=:black)
-vline!([mean(samples[1, :])]; color=:blue, label="")
+density!(samples[1, :]; ylabel="Density", label="s (ADVI)", color=:blue, linewidth=2)
+density!(samples_nuts, :s; ylabel="Density", label="s (NUTS)", color=:green, linewidth=2)
+vline!([var(x)]; ylabel="Density", label="s (data)", color=:black)
+vline!([mean(samples[1, :])]; ylabel="Density", color=:blue, label="")
```
devmotion, line 166: You don't have to set `ylabel` in every plotting call. It should be sufficient to use
```suggestion
p1 = histogram(samples[1, :]; bins=100, normed=true, alpha=0.2, color=:blue, label="", ylabel="density")
density!(samples[1, :]; label="s (ADVI)", color=:blue, linewidth=2)
density!(samples_nuts, :s; label="s (NUTS)", color=:green, linewidth=2)
vline!([var(x)]; label="s (data)", color=:black)
vline!([mean(samples[1, :])]; color=:blue, label="")
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -154,27 +155,23 @@ For that we need samples:
...
 
 p2 = histogram(samples[2, :]; bins=100, normed=true, alpha=0.2, color=:blue, label="")
-density!(samples[2, :]; label="m (ADVI)", color=:blue, linewidth=2)
-density!(samples_nuts, :m; label="m (NUTS)", color=:green, linewidth=2)
-vline!([mean(x)]; color=:black, label="m (data)")
-vline!([mean(samples[2, :])]; color=:blue, label="")
+density!(samples[2, :]; ylabel="Density", label="m (ADVI)", color=:blue, linewidth=2)
+density!(samples_nuts, :m; ylabel="Density", label="m (NUTS)", color=:green, linewidth=2)
+vline!([mean(x)]; ylabel="Density", color=:black, label="m (data)")
+vline!([mean(samples[2, :])]; ylabel="Density", color=:blue, label="")
```
devmotion, line 172: Same here:
```suggestion
p2 = histogram(samples[2, :]; bins=100, normed=true, alpha=0.2, color=:blue, label="", ylabel="density")
density!(samples[2, :]; label="m (ADVI)", color=:blue, linewidth=2)
density!(samples_nuts, :m; label="m (NUTS)", color=:green, linewidth=2)
vline!([mean(x)]; color=:black, label="m (data)")
vline!([mean(samples[2, :])]; color=:blue, label="")
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -273,28 +268,38 @@ end
...
-(std_data, data_arr) = standardize(Matrix(data))
-
 # Split our dataset 70%/30% into training/test sets.
-train, test = split_data(std_data, 0.7)
+train, test = split_data(data, 0.7)
+train_unstandardized = copy(train) #  save a copy of the train data for later unstandardize use; alternatively, you can just save the statistics (mean and var) of the traiing data.
+
+# Standardize both datasets.
+std_train = standardize(Matrix(train))
+std_test = standardize(Matrix(test), Matrix(train_unstandardized))
```
devmotion, line 298: Probably would be simpler to use existing functionality in some package but with the functions in the tutorial you could do
```suggestion

# Standardize both datasets.
std_train = standardize(Matrix(train))
std_test = standardize(Matrix(test), Matrix(train))
```
No need to copy anything since `train` is not updated in-place.

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -462,13 +481,18 @@ vi_mean = vec(mean(z; dims=2))[[
 ```
 
 ```julia
-mean(chain).nt.mean
+mcmc_mean = mean(chain).nt.mean
```
devmotion, line 484: This seems a bit strange, I don't think this is how users are supposed to compute the mean.

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -322,6 +327,10 @@ test = Matrix(test_cut[:, remove_names]);
 
     # Calculate all the mu terms.
     mu = intercept .+ x * coefficients
+    # for i in 1:n_obs
+    #     y[i] ~ Normal(mu[i], sqrt(σ²))
+    # end
+    # vectorized implementation
```
devmotion, line 333: IMO it's not good to add commented code. It will not be run and hence is not tested. Moreover, it increases the code complexity. If one wants to show an alternative, a separate model, sampling etc. would be better.

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -420,22 +432,29 @@ function plot_variational_marginals(z, sym2range)
...
-                    push!(ps, p)
-
-                    offset += 1
-                end
+                p = density(
+                    z[r, :];
+                    title="$(sym)[$offset]",
+                    titlefontsize=10,
+                    label="",
+                    margin=2.0mm,
```
devmotion, line 440: I'm a bit worried if the `margin` settings will be correct for all output formats (in HTML, markdown etc.).

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -462,13 +481,18 @@ vi_mean = vec(mean(z; dims=2))[[
 ```
 
 ```julia
-mean(chain).nt.mean
+mcmc_mean = mean(chain).nt.mean
+```
+
+```julia
+plot(mcmc_mean; xticks=1:1:length(mcmc_mean), linestyle=:dot, label="NUTS")
+plot!(vi_mean; xticks=1:1:length(mcmc_mean), linestyle=:dot, label="VI")
```
devmotion, line 489: Should not be necessary to set `xticks` twice:
```suggestion
plot!(vi_mean; linestyle=:dot, label="VI")
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -487,12 +511,12 @@ ols = lm(
 )
 
 # Store our predictions in the original dataframe.
-train_cut.OLSPrediction = unstandardize(GLM.predict(ols), data.MPG);
-test_cut.OLSPrediction = unstandardize(GLM.predict(ols, test_cut), data.MPG);
+train_cut.OLSPrediction = unstandardize(GLM.predict(ols), train_unstandardized.MPG);
```
devmotion, line 514:
```suggestion
train_cut.OLSPrediction = unstandardize(GLM.predict(ols), train.MPG)
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -487,12 +511,12 @@ ols = lm(
...
+test_cut.OLSPrediction = unstandardize(GLM.predict(ols, test_cut), train_unstandardized.MPG);
```
devmotion, line 515:
```suggestion
test_cut.OLSPrediction = unstandardize(GLM.predict(ols, test_cut), train.MPG);
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -518,8 +542,8 @@ end
 
 ```julia
 # Unstandardize the dependent variable.
-train_cut.MPG = unstandardize(train_cut.MPG, data.MPG);
-test_cut.MPG = unstandardize(test_cut.MPG, data.MPG);
+train_cut.MPG = unstandardize(train_cut.MPG, train_unstandardized.MPG);
+test_cut.MPG = unstandardize(test_cut.MPG, train_unstandardized.MPG);
```
devmotion, line 546:
```suggestion
train_cut.MPG = unstandardize(train_cut.MPG, train.MPG)
test_cut.MPG = unstandardize(test_cut.MPG, train.MPG);
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -533,11 +557,19 @@ z = rand(q, 10_000);
...
 ```julia
 # Calculate the predictions for the training and testing sets using the samples `z` from variational posterior
-train_cut.VIPredictions = unstandardize(prediction(z, sym2range, train), data.MPG);
-test_cut.VIPredictions = unstandardize(prediction(z, sym2range, test), data.MPG);
+train_cut.VIPredictions = unstandardize(
+    prediction(z, sym2range, train), train_unstandardized.MPG
+);
+test_cut.VIPredictions = unstandardize(
+    prediction(z, sym2range, test), train_unstandardized.MPG
+);
```
devmotion, line 565:
```suggestion
train_cut.VIPredictions = unstandardize(prediction(z, sym2range, train), train.MPG)
test_cut.VIPredictions = unstandardize(prediction(z, sym2range, test), train.MPG)
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -533,11 +557,19 @@ z = rand(q, 10_000);
...
 
-train_cut.BayesPredictions = unstandardize(prediction_chain(chain, train), data.MPG);
-test_cut.BayesPredictions = unstandardize(prediction_chain(chain, test), data.MPG);
+train_cut.BayesPredictions = unstandardize(
+    prediction_chain(chain, train), train_unstandardized.MPG
+);
+test_cut.BayesPredictions = unstandardize(
+    prediction_chain(chain, test), train_unstandardized.MPG
+);
```
devmotion, line 572:
```suggestion
train_cut.BayesPredictions = unstandardize(prediction_chain(chain, train), train.MPG)
test_cut.BayesPredictions = unstandardize(prediction_chain(chain, test), train.MPG);
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -564,7 +596,10 @@ Interestingly the squared difference between true- and mean-prediction on the te
 ```julia
 z = rand(q, 1000);
 preds = hcat(
-    [unstandardize(prediction(z[:, i], sym2range, test), data.MPG) for i in 1:size(z, 2)]...
+    [
+        unstandardize(prediction(z[:, i], sym2range, test), train_unstandardized.MPG) for
+        i in 1:size(z, 2)
+    ]...,
 );
```
devmotion, line 603: Better:
```suggestion
preds = mapreduce(hcat, eachcol(z)) do zi
    return unstandardize(prediction(zi, sym2range, test), train.MPG)
end
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -575,16 +610,16 @@ scatter(
     size=(900, 500),
     markersize=8,
 )
-scatter!(1:size(test, 1), unstandardize(test_label, data.MPG); label="true")
+scatter!(1:size(test, 1), unstandardize(test_label, train_unstandardized.MPG); label="true")
```
devmotion, line 613:
```suggestion
scatter!(1:size(test, 1), unstandardize(test_label, train.MPG); label="true")
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -575,16 +610,16 @@ scatter(
...
 ```
 
 ```julia
 preds = hcat(
     [
-        unstandardize(prediction_chain(chain[i], test), data.MPG) for
+        unstandardize(prediction_chain(chain[i], test), train_unstandardized.MPG) for
         i in 1:5:size(chain, 1)
     ]...,
 );
```
devmotion, line 625:
```suggestion
preds = mapreduce(hcat, 1:5:size(chain, 1)) do i
    return unstandardize(prediction_chain(chain[i], test), train.MPG)
end
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -597,15 +632,15 @@ scatter(
     size=(900, 500),
     markersize=8,
 )
-scatter!(1:size(test, 1), unstandardize(test_label, data.MPG); label="true")
+scatter!(1:size(test, 1), unstandardize(test_label, train_unstandardized.MPG); label="true")
```
devmotion, line 635:
```suggestion
scatter!(1:size(test, 1), unstandardize(test_label, train.MPG); label="true")
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -728,8 +763,12 @@ z = rand(q_full_normal, 10_000);
...
 
 ```julia
-train_cut.VIFullPredictions = unstandardize(prediction(z, sym2range, train), data.MPG);
-test_cut.VIFullPredictions = unstandardize(prediction(z, sym2range, test), data.MPG);
+train_cut.VIFullPredictions = unstandardize(
+    prediction(z, sym2range, train), train_unstandardized.MPG
+);
+test_cut.VIFullPredictions = unstandardize(
+    prediction(z, sym2range, test), train_unstandardized.MPG
+);
```
devmotion, line 771:
```suggestion
train_cut.VIFullPredictions = unstandardize(prediction(z, sym2range, train), train.MPG)
test_cut.VIFullPredictions = unstandardize(prediction(z, sym2range, test), train.MPG);
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -753,16 +792,13 @@ Test set:
     OLS loss: $ols_loss2")
 ```
 
-```julia; echo=false
-# Verify the loss on the test set.
-@assert vi_loss2 < 0.01
-@assert bayes_loss2 < 0.000001
-```
-
```
devmotion, line 761: Can we keep this? It is much easier to update and rerun tutorials with these checks (not visible for readers):
````suggestion
```julia; echo=false
# Verify the loss on the test set.
@assert vi_loss2 < 0.01
@assert bayes_loss2 < 0.000001
```

````

File: tutorials/09-variational-inference/09_variational-inference.jmd [hunk 37]
```diff
@@ -753,16 +792,13 @@ Test set:
...
 ```julia
 z = rand(q_mf_normal, 1000);
 preds = hcat(
-    [unstandardize(prediction(z[:, i], sym2range, test), data.MPG) for i in 1:size(z, 2)]...
+    [
+        unstandardize(prediction(z[:, i], sym2range, test), train_unstandardized.MPG) for
+        i in 1:size(z, 2)
+    ]...,
 );
```
devmotion, line 802:
```suggestion
preds = mapreduce(hcat, eachcol(z)) do zi
    return unstandardize(prediction(zi, sym2range, test), train.MPG)
end
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -773,16 +809,19 @@ p1 = scatter(
     size=(900, 500),
     markersize=8,
 )
-scatter!(1:size(test, 1), unstandardize(test_label, data.MPG); label="true")
+scatter!(1:size(test, 1), unstandardize(test_label, train_unstandardized.MPG); label="true")
```
devmotion, line 812:
```suggestion
scatter!(1:size(test, 1), unstandardize(test_label, train.MPG); label="true")
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -773,16 +809,19 @@ p1 = scatter(
...
 
 ```julia
 z = rand(q_full_normal, 1000);
 preds = hcat(
-    [unstandardize(prediction(z[:, i], sym2range, test), data.MPG) for i in 1:size(z, 2)]...
+    [
+        unstandardize(prediction(z[:, i], sym2range, test), train_unstandardized.MPG) for
+        i in 1:size(z, 2)
+    ]...,
 );
```
devmotion, line 825:
```suggestion
preds = mapreduce(hcat, eachcol(z)) do zi
   return unstandardize(prediction(zi, sym2range, test), train.MPG)
end
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -793,16 +832,16 @@ p2 = scatter(
     size=(900, 500),
     markersize=8,
 )
-scatter!(1:size(test, 1), unstandardize(test_label, data.MPG); label="true")
+scatter!(1:size(test, 1), unstandardize(test_label, train_unstandardized.MPG); label="true")
```
devmotion, line 835:
```suggestion
scatter!(1:size(test, 1), unstandardize(test_label, train.MPG); label="true")
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -793,16 +832,16 @@ p2 = scatter(
...
 ```
 
 ```julia
 preds = hcat(
     [
-        unstandardize(prediction_chain(chain[i], test), data.MPG) for
+        unstandardize(prediction_chain(chain[i], test), train_unstandardized.MPG) for
         i in 1:5:size(chain, 1)
     ]...,
 );
```
devmotion, line 847:
```suggestion
preds = mapreduce(hcat, 1:5:size(chain, 1)) do i
    return unstandardize(prediction_chain(chain[i], test), train.MPG)
end
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -815,9 +854,9 @@ p3 = scatter(
     size=(900, 500),
     markersize=8,
 )
-scatter!(1:size(test, 1), unstandardize(test_label, data.MPG); label="true")
+scatter!(1:size(test, 1), unstandardize(test_label, train_unstandardized.MPG); label="true")
```
devmotion, line 857:
```suggestion
scatter!(1:size(test, 1), unstandardize(test_label, train.MPG); label="true")
```

tutorials/09-variational-inference/09_variational-inference.jmd
devmotion:
```suggestion
```

tutorials/09-variational-inference/Project.toml
```diff
@@ -2,11 +2,12 @@
 Bijectors = "76274a88-744f-5084-9051-94815aaf08c4"
 ComponentArrays = "b0b7db55-cfe3-40fc-9ded-d10e2dbeff66"
 FillArrays = "1a297f60-69ca-5386-bcde-b61e274b549b"
+Flux = "587475ba-b771-5e3f-ad9e-33799f191a9c"
```
devmotion, line 5: I think Flux is not used anywhere?
```suggestion
```
(requires to rebuild Manifest.toml)

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: @devmotion Yes it will have to be tested. Locally `margin=2.00mm` looks good, but once rendered as HTML it normally requires `margin=6.00mm`, as found in the pPCA tutorial.
I just changed it to be `margin=6.00mm` to see how it works - maybe revert it later.

tutorials/09-variational-inference/Project.toml
YongchaoHuang: Great, `ConjugatePriors` was removed and `p_μ_pdf ` reversed.

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: Cool, the `mapreduce...do...` is much simpler!

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: Good idea. So `echo=false` sets no output but will evaluate the block; `evaluate=False` will skip the evaluation?

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: yes, it works - I just reverted it to:
```p_μ = μₙ + sqrt(βₙ / (αₙ * κₙ)) * TDist(2 * αₙ)
p_μ_pdf = z -> pdf(p_μ, z)
```

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: True - just modified it.

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: Yes, it simplifies the codes.

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: yes agree, we can use `chain[:intercept]` etc to extract the sample trajectories, however, we have 13 variables and it would be less elegant to do so. Any quick way maybe?

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: yes, just modified.

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: yes, suggestions applied.

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: yes, suggestions applied.

tutorials/09-variational-inference/09_variational-inference.jmd
devmotion: It also also much more efficient, in particular for large arrays: Generally one should not try to splat large arrays.

tutorials/09-variational-inference/09_variational-inference.jmd
devmotion: I don't remember all options but yes, `echo=false` ensures that nothing is printed but the code is evaluated.

tutorials/09-variational-inference/Project.toml
YongchaoHuang: @devmotion I guess it's used in Line 362: 
```
vi(m, advi; optimizer=Flux.ADAM())
```
as mentioned in Line 356: "With that being said, thanks to Requires.jl, if we add a `using Flux` prior to `using Turing` we can also make use of all the optimizers in `Flux`, e.g. `ADAM`, without any additional changes to your code! For example ..."

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: Great to know. So the savings are mainly due to `mapreduce`?

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: @devmotion Actually `train` is over-written later as a Matrix in Line 311:
```
train = Matrix(train_cut[:, remove_names]);
test = Matrix(test_cut[:, remove_names]);
```
This is from the original version and I therefore reverted the use of `train_unstandardized` here.

YongchaoHuang:
@devmotion Thanks a lot for your help in improving this PR, the build is successful (see [here](https://julialang-buildkite-artifacts.s3.amazonaws.com/turingtutorials/807f4def0960696a046027ef697ef3eea3b6ee0b88e22ace8321393846b26a49/html/09-variational-inference/09_variational-inference.html)). 
Maybe ready to be merged?

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -154,27 +155,26 @@ For that we need samples:
...
 vline!([mean(samples[1, :])]; color=:blue, label="")
 
-p2 = histogram(samples[2, :]; bins=100, normed=true, alpha=0.2, color=:blue, label="")
+p2 = histogram(
+    samples[2, :]; bins=100, normed=true, alpha=0.2, color=:blue, label="", ylabel="density"
+)
 density!(samples[2, :]; label="m (ADVI)", color=:blue, linewidth=2)
 density!(samples_nuts, :m; label="m (NUTS)", color=:green, linewidth=2)
 vline!([mean(x)]; color=:black, label="m (data)")
-vline!([mean(samples[2, :])]; color=:blue, label="")
```
devmotion, line 175: Was this line removed on purpose?

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -395,7 +403,10 @@ _, sym2range = bijector(m, Val(true));
 sym2range
 ```
 
+For example, we can check the sample distribution and mean value of `σ²`
```
devmotion, line 406:
```suggestion
For example, we can check the sample distribution and mean value of `σ²`:
```

tutorials/09-variational-inference/09_variational-inference.jmd
devmotion: Maybe something like
```suggestion
mcmc_mean = mean(chain, names(chain, :parameters))
```
(untested)

tutorials/09-variational-inference/09_variational-inference.jmd
devmotion: Due to avoiding splatting and since mapreduce has optimized implementations for hcat.
BTW it seems the suggestion was not applied?

tutorials/09-variational-inference/Project.toml
devmotion: Ah Isee!

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: Sorry, it's a mistakenly removed. Recovered now.

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: hmm...strange, I did apply all of the suggestions. Now should be applied.

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: Great, thanks, I got it. Two options: `echo` and `eva` for the code blocks.

tutorials/09-variational-inference/09_variational-inference.jmd
devmotion: Did you forget to push your changes? It seems it's still missing on Github.

tutorials/09-variational-inference/09_variational-inference.jmd
devmotion: Still missing on Github it seems.

tutorials/09-variational-inference/09_variational-inference.jmd
devmotion: The assertions are not recovered on Github yet.

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: Yes, making the commit soon.

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: Yes, tested as follows:
```
mean(chain, names(chain, :parameters))[:, 2]
```

tutorials/09-variational-inference/09_variational-inference.jmd
YongchaoHuang: sorry I was finally testing all blocks - just finished to make a overall commit.

YongchaoHuang:
@devmotion I applied most (if not all) of the changes suggested, but get messed up when replying to your suggestions on web and VScode, and some replies are not shown (now sure why). The most recent commit `d221e21` should have all changes applied. Ready to merge? Thanks.

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -154,27 +155,26 @@ For that we need samples:
...
+p2 = histogram(
+    samples[2, :]; bins=100, normed=true, alpha=0.2, color=:blue, label="", ylabel="density"
+)
 density!(samples[2, :]; label="m (ADVI)", color=:blue, linewidth=2)
 density!(samples_nuts, :m; label="m (NUTS)", color=:green, linewidth=2)
 vline!([mean(x)]; color=:black, label="m (data)")
 vline!([mean(samples[2, :])]; color=:blue, label="")
-
-plot(p1, p2; layout=(2, 1), size=(900, 500))
+plot(p1, p2; layout=(2, 1), size=(900, 500), legend=true)
```
devmotion, line 177:
```suggestion

plot(p1, p2; layout=(2, 1), size=(900, 500), legend=true)
```

tutorials/09-variational-inference/09_variational-inference.jmd
```diff
@@ -753,17 +787,11 @@ Test set:
     OLS loss: $ols_loss2")
 ```
 
-```julia; echo=false
-# Verify the loss on the test set.
-@assert vi_loss2 < 0.01
-@assert bayes_loss2 < 0.000001
-```
-
 ```julia
```
devmotion, line 790:
````suggestion
```julia; echo=false
# Verify the loss on the test set.
@assert vi_loss2 < 0.01
@assert bayes_loss2 < 0.000001
```

```julia
````

devmotion (approved):
Looks good, I pushed a final commit that recovered some lines that (I assume) were removed accidentally but not recovered. Just check if the plots and terminal outputs look good.

YongchaoHuang:
> Looks good, I pushed a final commit that recovered some lines that (I assume) were removed accidentally but not recovered. Just check if the plots and terminal outputs look good.

I am adjusting the margin in `plot(chain; margin=4.00mm)` - making a final commit.

yebai:
Excellent improvements @YongchaoHuang @devmotion!

PR #350, opened 2022-11-24
Update 08_multinomial-logistic-regression.jmd

YongchaoHuang:
corrected a descriptive sentence:
`How do we test how well the model actually predicts whether someone is likely to default?` (not the logistic student default example from ISLR) to
`How do we test how well the model actually predicts which of the three classes an iris flower belongs to?`


yebai (approved):
thanks, @YongchaoHuang!

PR #351, opened 2022-11-24
Update 10_bayesian-differential-equations.jmd

YongchaoHuang:
suggested new ICs for simulation; 
suggested using real-world LV data.


tutorials/10-bayesian-differential-equations/10_bayesian-differential-equations.jmd
```diff
@@ -56,7 +56,7 @@ function lotka_volterra(du, u, p, t)
 end
 
 # Define initial-value problem.
-u0 = [1.0, 1.0]
+u0 = [1.0, 1.0] # might be more reasonable to use the initial conditions u0 = [1.0, 0.0]
```
yebai, line 59:
```suggestion
u0 = [1.0, 0.] 
```

tutorials/10-bayesian-differential-equations/10_bayesian-differential-equations.jmd
```diff
@@ -78,6 +78,12 @@ plot(sol; alpha=0.3)
 scatter!(sol.t, odedata'; color=[1 2], label="")
 ```
 
+# alternatively, we can use real-world data from Hudson’s Bay Company records (a Stan implementation can be found here: https://mc-stan.org/users/documentation/case-studies/lotka-volterra-predator-prey.html)
```
yebai, line 81: @ChrisRackauckas would that be a good idea?

ChrisRackauckas: yeah seems fine

tutorials/10-bayesian-differential-equations/10_bayesian-differential-equations.jmd
```diff
@@ -78,6 +78,19 @@ plot(sol; alpha=0.3)
...
+# years = [1900,1901,1902,1903,1904,1905,1906,1907,1908,1909,1910,1911,1912,1913,1914,1915,1916,1917,1918,1919,1920]
+
+# prey
+
+hare = [30.0, 47.2, 70.2, 77.4, 36.3, 20.6, 18.1, 21.4, 22.0, 25.4, 27.1, 40.3, 57.0, 76.6, 52.3, 19.5, 11.2, 7.6, 14.6, 16.2, 24.7]
+
+# predator
+
+lynx = [4.0, 6.1, 9.8, 35.2, 59.4, 41.7, 19.0, 13.0, 8.3, 9.1, 7.4, 8.0, 12.3, 19.5, 45.7, 51.1, 29.7, 15.8, 9.7, 10.1, 8.6]
+odedata = [hare lynx]'
```
devmotion, line 92: This should be put inside of a code block, shouldn't it? Or we should only add the comment?

YongchaoHuang:
@ChrisRackauckas @yebai I just added the real-world data from the Stan tutorial (https://mc-stan.org/users/documentation/case-studies/lotka-volterra-predator-prey.html), assigned it to `odedata` as an alternative to the simulated data. 
Should be ready to be merged if you think it's appropriate. Thanks.

tutorials/10-bayesian-differential-equations/10_bayesian-differential-equations.jmd
YongchaoHuang: > This should be put inside of a code block, shouldn't it? Or we should only add the comment?

@devmotion Good point, yes it should sit in the `julia` block. Just fixed it :)

tutorials/10-bayesian-differential-equations/10_bayesian-differential-equations.jmd
```diff
@@ -73,6 +73,63 @@ To make the example more realistic we add random normally distributed noise to t
 sol = solve(prob, Tsit5(); saveat=0.1)
 odedata = Array(sol) + 0.8 * randn(size(Array(sol)))
 
+# alternatively, we can use real-world data from Hudson’s Bay Company records (a Stan implementation can be found here: https://mc-stan.org/users/documentation/case-studies/lotka-volterra-predator-prey.html)
```
devmotion, line 76: As it is written currently, it is not an alternative: It is used instead of the simulated data. So the explanation above would have to be fixed and the code that generates the noise simulated data could be removed.

tutorials/10-bayesian-differential-equations/10_bayesian-differential-equations.jmd
```diff
@@ -73,6 +73,63 @@ To make the example more realistic we add random normally distributed noise to t
...
+
+# years = [1900,1901,1902,1903,1904,1905,1906,1907,1908,1909,1910,1911,1912,1913,1914,1915,1916,1917,1918,1919,1920]
```
devmotion, line 78: We need also time points, otherwise we don't know how to solve the ODE.

YongchaoHuang: > We need also time points, otherwise we don't know how to solve the ODE.

@devmotion that's true, thanks for the reminder. I just modified `tspan` to accommodate the new data.

tutorials/10-bayesian-differential-equations/10_bayesian-differential-equations.jmd
YongchaoHuang: @devmotion yes it's replacing the simulated data. Maybe we can keep both?

devmotion: IMO that's just confusing. I think either we present both approaches fully or we just remove the dead code. I think I'm in favour of the latter, to keep the size of the tutorial and its execution times manageable.

Note though that Stan uses a different model for the Hudson Bay dataset (different priors etc.). So I'm not sure at all that the different dataset will produce useful results at all currently (or rather I assume it doesn't).

YongchaoHuang: @devmotion Thanks David for offering the thoughts, yes, I just re-removed the real-world data because:
1. leaving the choice to users may be confusing;
2. The Stan model is a bit different from ours here. 
as you mentioned in previous comments.

devmotion (approved):
Looks good to me, seems quite uncontroversial to me :slightly_smiling_face:  (I think the Stan write-up is a very nice BTW.)

YongchaoHuang:
> 

@devmotion Thanks very much for your time and useful advice. Yes I also found the [Stan](https://mc-stan.org/users/documentation/case-studies/lotka-volterra-predator-prey.html), [Tensorflow](https://www.tensorflow.org/probability/examples/Probabilistic_PCA), and [Edward/Python](http://edwardlib.org/tutorials/probabilistic-pca) p-PCA tutorials useful.

yebai:
thanks @YongchaoHuang!

PR #369, opened 2022-12-22
Some fixes

devmotion:
Some fixes of CI originally part of #368.

### DIFF
--- a/.buildkite/0_webui.yml
+++ b/.buildkite/0_webui.yml
@@ -1,6 +1,6 @@
 agents:
   queue: "juliaecosystem"
-  sandbox.jl: true
+  sandbox_capable: true
 
 steps:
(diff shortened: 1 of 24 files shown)


yebai (approved):

devmotion:
@staticfloat Sorry to bother you again: On Slack I read that `sandbox.jl` was replaced with `sandbox_capable` and I updated the pipeline accordingly (and indeed now tutorials are built again whereas previously they were just stuck in the queue forever). Unfortunately, that means that the tree hash has to be updated again (so far I've unblocked the pipeline manually) - could you push the necessary changes (IRC it's not something I can do on my own)?

staticfloat:
Done, and pushed straight to master :)

devmotion:
Thank you!

PR #372, opened 2023-01-01
Replace MLDataUtils with MLUtils

### LINKED ISSUE #322: Question: Update from MLDataUtils to MLUtils
Question on: https://github.com/TuringLang/Turing.jl/edit/master/docs/_tutorials/05_linear-regression.md
This tutorial makes use of MLDataUtils, which is archived and now read-only. The readme for MLDataUtils states that
"This package has been discontinued. Most functionalities have been moved [MLUtils.jl](https://github.com/JuliaML/MLUtils.jl)."

This issue is just a to-do item to update to using MLUtils, in order to avoid pointing new users towards depricated packages.

devmotion:
Fixes #322.


tutorials/05-linear-regression/05_linear-regression.jmd
```diff
@@ -75,12 +74,14 @@ train_target = trainset[:, target]
...
-μ, σ = rescale!(train; obsdim=1)
-rescale!(test, μ, σ; obsdim=1)
+dt_features = fit(ZScoreTransform, train; dims=1)
+StatsBase.transform!(dt_features, train)
+StatsBase.transform!(dt_features, test)
 
 # Standardize the targets.
-μtarget, σtarget = rescale!(train_target; obsdim=1)
-rescale!(test_target, μtarget, σtarget; obsdim=1);
+dt_targets = fit(ZScoreTransform, train_target)
```
YongchaoHuang, line 82: Personally a bit curious about the difference between `MLDataUtils.rescale!` and `StatsBase.fit and transform`: are they producing the same results (i.e. mean and std)? In R and Python they are the same, e.g. in Scikit-learn pre-processing module it's called scalers (normal, min-max, etc) with methods `fit`, `transform` or `fit_transform`.

tutorials/05-linear-regression/05_linear-regression.jmd
```diff
@@ -153,16 +154,14 @@ using GLM
 train_with_intercept = hcat(ones(size(train, 1)), train)
 ols = lm(train_with_intercept, train_target)
 
-# Compute predictions on the training data set
-# and unstandardize them.
-p = GLM.predict(ols)
-train_prediction_ols = μtarget .+ σtarget .* p
+# Compute predictions on the training data set and unstandardize them.
+train_prediction_ols = GLM.predict(ols)
+StatsBase.reconstruct!(dt_targets, train_prediction_ols)
```
YongchaoHuang, line 159: so the inner working of `StatsBase.reconstruct` is: it first calculates the mean and std from the training targets, and then apply them to the predicted (scaled) targets?

tutorials/05-linear-regression/05_linear-regression.jmd
```diff
@@ -65,7 +64,7 @@ The next step is to get our data ready for testing. We'll split the `mtcars` dat
 select!(data, Not(:Model))
 
 # Split our dataset 70%/30% into training/test sets.
-trainset, testset = splitobs(shuffleobs(data), 0.7)
+trainset, testset = map(DataFrame, splitobs(data; at=0.7, shuffle=true))
```
YongchaoHuang, line 67: Great to know the `splitobs` has an argument `shuffle` - it looks quite similar to `sklearn.model_selection.train_test_split(shuffle=True)`.

tutorials/05-linear-regression/05_linear-regression.jmd
devmotion: There are many more alternwtives (eg in TableTransforms or FeatureTransforms) but StatsBase is sufficient here and a indirect dependency anyway. It yields the same result as the previous version with the deprecated MLDataUtils (this functionality was not and will not be transferred to MLUtils), but the version in StatsBase only works with vectors and matrices (which is sufficient here).

tutorials/05-linear-regression/05_linear-regression.jmd
devmotion: No. You compute a transformation object with `fit` which in the case of `ZScoreTransform` mainly stores empirical mean and standard deviation/variance of each feature. Then you can apply this transformation with `transform`/`transform!` which in this case will normalize the data. And with `reconstruct`/`reconstruct!` you can un-normalize data by scaling it with the learnt standard deviation and shifting it with the mean.

tutorials/05-linear-regression/05_linear-regression.jmd
devmotion: There are many other alternatives. I guess if you like scikit-learn (I only use it when I'm forced to for teaching 😉), then you might prefer MLJ.

YongchaoHuang: > There are many other alternatives. I guess if you like scikit-learn (I only use it when I'm forced to for teaching wink), then you might prefer MLJ.

Thanks a lot for mentioning MLJ - it's a very interesting interface. Yes I used sklearn/tidyverse/dplyr in the early days back to 17/18 for quick model deployment, so feel quite curious about similar syntax in Julia. Thanks for the illustrations, very well demonstrated.

YongchaoHuang (commented):
@devmotion Thanks very much for greatly simplifying the packages and optimizing the methods.

YongchaoHuang (approved):

yebai (approved):

PR #375, opened 2023-01-06
Transfer all high level docs from Turing.jl

yebai:
This PR transfers all high-level (i.e. non-API or doc-string) documentation from `Turing.jl`. This allows us to unify most docs and tutorials workflow and pave the way to simplify build steps (we only need standard Jekyll eventually) in the `turing.ml` repo. 

TODO
 - convert all `md` files into `jmd` files.


yebai:
@cgbotta I added default project files to all docs. The default environment only installs `Turing` and `StatsPlots`. This is likely insufficient for more complicated doc pages. Please feel free to add additional dependencies where needed.

cgbotta:
When I try to weave the new tutorials using `TuringTutorials.weave` it only weaves the existing ones, but does not include the new ones we have added in this PR

I recall that there was one place where I needed to manually add new tutorials to the weave scripts, but this list seems to have been removed in @devmotion 's recent work. Am I overlooking something?

tutorials/using-turing/index.jmd
```diff
@@ -0,0 +1,20 @@
+---
```
devmotion, line 1: This file does not contain any code or examples but is just a landing page on the webpage it seems, so I wonder:
Should this file be added here? Or should it be a regular markdown file? Or at least, do we really have to add Project.toml  + Manifest.toml files here?

devmotion:
> When I try to weave the new tutorials using `TuringTutorials.weave` it only weaves the existing ones, but does not include the new ones we have added in this PR
> 
> I recall that there was one place where I needed to manually add new tutorials to the weave scripts, but this list seems to have been removed in @devmotion 's recent work. Am I overlooking something

I guess you might refer to the manual list in the CompatHelper github action? The whole action was removed and replaced with an action that updates all Manifest.toml files (run once every week), so you're right, you don't have to add new tutorials/docs to some list anymore. However, that list was only relevant for CI (or more specifically CompatHelper) and completely independent from the `weave` commands (and basically everything else in the repo).

There were no changes to the `weave` commands after [your PR](https://github.com/TuringLang/TuringTutorials/pull/327) it seems.

Regarding the new jmd files in this PR, it seems in the previous PR the non-tutorial jmd file was put in the tutorials folder but with a `docs-` prefix. Should we also somehow make it clear (e.g. by using a prefix) that these new jmd files are distinct from the tutorials, even though they are in the tutorials folder? Or maybe it would be better to move non-tutorials to a separate folder (docs?)?

cgbotta:
I spent a while fixing all the BuiltKite errors and generating the missing artifacts. There are just two that are behaving weirdly. Could either of you let me know if you are able to fix these ones? @yebai @devmotion 

- using-turing-guide cannot find Turing, which I was able to fix on other tutorials, but I have no idea why this one keeps failing
- autodiff is failing on an internal function

devmotion:
I think this PR should not have been merged - I really think we should fix #394 before any other changes in this repo. Since it is not resolved all the docs and all the examples will be built and run with Julia 1.6 - but since Turing > 0.22 is not compatible with Julia 1.6 anymore, docs will be generated with a quite old version of Turing and in particular are not representative of the latest release (the versioning on the webpage has been misleading for a long time, even though the tutorials are listed as /v26/ they do *not* use Turing 0.26).

PR #377, opened 2023-01-07
Yongchao/p pca fix rendering

YongchaoHuang:
Try to fix the MathJax rendering problem.

### DIFF
--- a/tutorials/11-probabilistic-pca/11_probabilistic-pca.jmd
+++ b/tutorials/11-probabilistic-pca/11_probabilistic-pca.jmd
@@ -14,7 +14,7 @@ It is an unsupervised learning method mainly used for dimensionality reduction.
 For example, we have a data matrix $\textbf{X} \in \mathbb{R}^{N \times D}$, and we would like to extract $k \ll D$ principal components which captures most of the information from the original matrix.
 The goal is to understand $\textbf{X}$ through a lower dimensional subspace (e.g. two-dimensional subspace for visualisation convenience) spanned by the principal components.
 
-In order to project the original data matrix into low dimensions, we need to find the principal directions where most of the variations of $\mathbf{X}$ lie in.
+In order to project the original data matrix into low dimensions, we need to find the principal directions where most of the variations of $\textbf{X}$ lie in.
 Traditionally, this is implemented via [singular value decomposition (SVD)](https://en.wikipedia.org/wiki/Singular_value_decomposition) which provides a robust and accurate computational framework for decomposing matrix into products of rotation-scaling-rotation matrices, particularly for large datasets(see an illustration [here](https://intoli.com/blog/pca-and-svd/)):
 
 $$
@@ -48,13 +48,13 @@ $$
 and the corresponding data point is generated via projection:
 
 $$
-x_i | z_i \sim \mathcal{N}(\textbf{W} z_i + \boldsymbol{μ}, \sigma^2 \textbf{I})
+x_i | z_i \sim \mathcal{N}(\textbf{W} z_i + \textbf{\mu}, \sigma^2 \textbf{I})
 $$
 
 where the projection matrix $\textbf{W}_{D \times k}$ accommodates the principal axes.
-The above formula expresses $x_i$ as a linear combination of the basis columns in the projection matrix `W`, where the combination coefficients sit in `z_i` (they are the coordinats of `x_i` in the new $k$-dimensional space.).
+The above formula expresses $x_i$ as a linear combination of the basis columns in the projection matrix $\textbf{W}$, where the combination coefficients sit in $z_i$ (they are the coordinats of $x_i$ in the new $k$-dimensional space.).
 We can also express the above formula in matrix form: $\textbf{X}_{D \times N} \approx \textbf{W}_{D \times k} \textbf{Z}_{k \times N}$.
-We are interested in inferring $\textbf{W}$, $μ$ and $\sigma$.
+We are interested in inferring $\textbf{W}$, $\mu$ and $\sigma$.
 
 Classical PCA is the specific case of probabilistic PCA when the covariance of the noise becomes infinitesimally small, i.e. $\sigma^2 \to 0$.
 Probabilistic PCA generalizes classical PCA, this can be seen by marginalizing out the the latent variable.[^2]
@@ -64,7 +64,7 @@ Probabilistic PCA generalizes classical PCA, this can be seen by marginalizing o
 In the first example, we illustrate:
 
   - how to specify the probabilistic model and
-  - how to perform inference on $\textbf{W}$, $\boldsymbol{\mu}$ and $\sigma$ using MCMC.
+  - how to perform inference on $\textbf{W}$, $\textbf{\mu}$ and $\sigma$ using MCMC.
 
 We use simulated gemnome data to demonstrate these.
 The simulation is inspired by biological measurement of expression of genes in cells, and each cell is characterized by different gene features.
@@ -151,7 +151,7 @@ Note that:
 #### Step 3: Create the pPCA model
 
 Here we construct the probabilistic model `pPCA()`.
-As per the p-PCA formula, we think of each row (i.e. each gene feature) following a $N=60$ dimensional multivariate normal distribution centered around the corresponding row of $\mathbf{W}_{D \times k} \times \mathbf{Z}_{k \times N} + \boldsymbol{\mu}_{D \times N}$.
+As per the p-PCA formula, we think of each row (i.e. each gene feature) following a $N=60$ dimensional multivariate normal distribution centered around the corresponding row of $\textbf{W}_{D \times k} \times \textbf{Z}_{k \times N} + \textbf{\mu}_{D \times N}$.
 
 ```julia
 @model function pPCA(X::AbstractMatrix{<:Real}, k::Int)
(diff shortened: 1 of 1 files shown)


tutorials/11-probabilistic-pca/11_probabilistic-pca.jmd
```diff
@@ -11,32 +11,32 @@ weave_options:
 Principal component analysis (PCA) is a fundamental technique to analyse and visualise data.
 It is an unsupervised learning method mainly used for dimensionality reduction.
 
-For example, we have a data matrix $\mathbf{X} \in \mathbb{R}^{N \times D}$, and we would like to extract $k \ll D$ principal components which captures most of the information from the original matrix.
-The goal is to understand $\mathbf{X}$ through a lower dimensional subspace (e.g. two-dimensional subspace for visualisation convenience) spanned by the principal components.
+For example, we have a data matrix $\textbf{X} \in \mathbb{R}^{N \times D}$, and we would like to extract $k \ll D$ principal components which captures most of the information from the original matrix.
```
devmotion, line 14: AFAIK it's not correct to use `textbf` here. IIRC `\mathbf{X}` also works fine already, doesn't it?

tutorials/11-probabilistic-pca/11_probabilistic-pca.jmd
```diff
@@ -11,32 +11,32 @@ weave_options:
...
+The goal is to understand $\textbf{X}$ through a lower dimensional subspace (e.g. two-dimensional subspace for visualisation convenience) spanned by the principal components.
 
-In order to project the original data matrix into low dimensions, we need to find the principal directions where most of the variations of $\mathbf{X}$ lie in.
+In order to project the original data matrix into low dimensions, we need to find the principal directions where most of the variations of $\textbf{X}$ lie in.
 Traditionally, this is implemented via [singular value decomposition (SVD)](https://en.wikipedia.org/wiki/Singular_value_decomposition) which provides a robust and accurate computational framework for decomposing matrix into products of rotation-scaling-rotation matrices, particularly for large datasets(see an illustration [here](https://intoli.com/blog/pca-and-svd/)):
 
 $$
-\mathbf{X}_{N \times D} =  \mathbf{U}_{N \times r} \times \boldsymbol{\Sigma}_{r \times r}  \times  \mathbf{V}^T_{r \times D}
+\textbf{X}_{N \times D} =  \textbf{U}_{N \times r} \times \textbf{Σ}_{r \times r}  \times  \textbf{V}^T_{r \times D}
```
devmotion, line 21: I think `\boldsymbol` might be the only problematic part here. Also should use `\Sigma`.

tutorials/11-probabilistic-pca/11_probabilistic-pca.jmd
YongchaoHuang: Yes, both `textbf` and `mathbf` give bold texts, `mathbf` sounds more evidenced but not sure if it's rendered. let me change one thing each time and see how it works.

tutorials/11-probabilistic-pca/11_probabilistic-pca.jmd
YongchaoHuang: Agree, but with `Sigma` we need `boldsymbol`, while we can use `mathbf{Σ} to avoid `boldsymbol`?

YongchaoHuang:
@devmotion The most recent version works well (see [the html file] (https://julialang-buildkite-artifacts.s3.amazonaws.com/turingtutorials/64b55d5ac9616d9a9bd35a0df65e189e6dc59c0053a3ed647a3ca14d59d80683/html/11-probabilistic-pca/11_probabilistic-pca.html). Shall we merge it?

devmotion:
I don't think the changes should be applied. Most LaTeX distributions don't work well with unicode characters, we should just use `\Sigma` and `\mu` as in other tutorials. Additionally, the HTML output is not a good indication for whether this actually fixes the issues on the webpage since that  is not based on the HTML output.

Maybe we do not have to fix the tutorial at all but only enable the boldsymbol extension on the homepage (ie in the turing.ml repo): https://docs.mathjax.org/en/latest/input/tex/extensions/boldsymbol.html

YongchaoHuang:
> I don't think the changes should be applied. Most LaTeX distributions don't work well with unicode characters, we should just use `\Sigma` and `\mu` as in other tutorials. Additionally, the HTML output is not a good indication for whether this actually fixes the issues on the webpage since that is not based on the HTML output.
> 
> Maybe we do not have to fix the tutorial at all but only enable the boldsymbol extension on the homepage (ie in the turing.ml repo): https://docs.mathjax.org/en/latest/input/tex/extensions/boldsymbol.html

@devmotion Many thanks for the knowledge sharing - yes we may only need to fix the `boldsymbol`. A qucik solution, as suggested in the link, is to dynamically load it via `\require{boldsymbol}`. We can test this solution to see if the problem stems from `boldsymbol`.
Shall I delete this PR and create a new one to just add `\require{boldsymbol}'?

devmotion:
I just checked and the boldsymbol extension is already loaded. Based on the source code of the webpage (https://turing.ml/v0.23/tutorials/11-probabilistic-pca/) the problem seems to be how Jekyll and its plugins parse the underscores: For instance, the first problematic equation shows up as
```html
$$
\mathbf{X}<em>{N \times D} =  \mathbf{U}</em>{N \times r} \times \boldsymbol{\Sigma}<em>{r \times r}  \times  \mathbf{V}^T</em>{r \times D}
$$
```

It seems this is the same problem as discussed in https://github.com/TuringLang/Turing.jl/pull/1869#discussion_r929563595. So possibly actually the correct fix is escaping the underscores.

YongchaoHuang:
> lly the corr



> I just checked and the boldsymbol extension is already loaded. Based on the source code of the webpage (https://turing.ml/v0.23/tutorials/11-probabilistic-pca/) the problem seems to be how Jekyll and its plugins parse the underscores: For instance, the first problematic equation shows up as
> 
> ```
> $$
> \mathbf{X}<em>{N \times D} =  \mathbf{U}</em>{N \times r} \times \boldsymbol{\Sigma}<em>{r \times r}  \times  \mathbf{V}^T</em>{r \times D}
> $$
> ```
> 
> It seems this is the same problem as discussed in [TuringLang/Turing.jl#1869 (comment)](https://github.com/TuringLang/Turing.jl/pull/1869#discussion_r929563595). So possibly actually the correct fix is escaping the underscores.

That seems quite possible, as in both equations there are multiple underscores and both of them are not correctly rendered.
I will just modify it to be `\_`.

devmotion:
Seems now the HTML output is broken: https://julialang-buildkite-artifacts.s3.amazonaws.com/turingtutorials/3f0c4fc98ee1135835a4a279bb2bd4b48b0efe443e56665e932902eda47f6e17/html/11-probabilistic-pca/11_probabilistic-pca.html

I think we should try to figure why Jekyll and its plugin have a problem with the underscores. Maybe we can fix the issue in a different way once we understand it better.

YongchaoHuang:
> Seems now the HTML output is broken: https://julialang-buildkite-artifacts.s3.amazonaws.com/turingtutorials/3f0c4fc98ee1135835a4a279bb2bd4b48b0efe443e56665e932902eda47f6e17/html/11-probabilistic-pca/11_probabilistic-pca.html
> 
> I think we should try to figure why Jekyll and its plugin have a problem with the underscores. Maybe we can fix the issue in a different way once we understand it better.

Yes, because the `\_` is recoginised as an symbol `_` in Julia; but it should work on Turing.ml, as explained in the issue #1869.
How about the following solution?
Maybe we also need to include that line in _layouts/post.html? as suggested [here](http://www.iangoodfellow.com/blog/jekyll/markdown/tex/2016/11/07/latex-in-markdown.html).
It suggests [here](https://janhendrikewers.uk/adding_latex_to_jekyll.html) and [here](https://stackoverflow.com/questions/51015485/jekyll-latex-not-showing-up-in-remote-but-shows-up-locally)  that, [cdn.mathjax.org](http://cdn.mathjax.org/) has been deprecated. Please use [cdnjs.cloudflare.com](http://cdnjs.cloudflare.com/) .

devmotion:
> Maybe we also need to include that line in _layouts/post.html? as suggested [here](http://www.iangoodfellow.com/blog/jekyll/markdown/tex/2016/11/07/latex-in-markdown.html).
> It suggests [here](https://janhendrikewers.uk/adding_latex_to_jekyll.html) and [here](https://stackoverflow.com/questions/51015485/jekyll-latex-not-showing-up-in-remote-but-shows-up-locally) that, [cdn.mathjax.org](http://cdn.mathjax.org/) has been deprecated. Please use [cdnjs.cloudflare.com](http://cdnjs.cloudflare.com/) .

I'm a bit confused, we don't use cdn.mathjax.org, do we? We use https://github.com/TuringLang/turing.ml/blob/996b5298999190b6582f98a3aeb7240c982a0b9e/_includes/headers.html#L44.

I'm a bit surprised about this line though: https://github.com/TuringLang/turing.ml/blob/996b5298999190b6582f98a3aeb7240c982a0b9e/_includes/headers.html#L39 Double-dollar's shouldn't be parsed as inline math, should they?

YongchaoHuang:
> 



> > Maybe we also need to include that line in _layouts/post.html? as suggested [here](http://www.iangoodfellow.com/blog/jekyll/markdown/tex/2016/11/07/latex-in-markdown.html).
> > It suggests [here](https://janhendrikewers.uk/adding_latex_to_jekyll.html) and [here](https://stackoverflow.com/questions/51015485/jekyll-latex-not-showing-up-in-remote-but-shows-up-locally) that, [cdn.mathjax.org](http://cdn.mathjax.org/) has been deprecated. Please use [cdnjs.cloudflare.com](http://cdnjs.cloudflare.com/) .
> 
> I'm a bit confused, we don't use cdn.mathjax.org, do we? We use https://github.com/TuringLang/turing.ml/blob/996b5298999190b6582f98a3aeb7240c982a0b9e/_includes/headers.html#L44.
> 
> I'm a bit surprised about this line though: https://github.com/TuringLang/turing.ml/blob/996b5298999190b6582f98a3aeb7240c982a0b9e/_includes/headers.html#L39 Double-dollar's shouldn't be parsed as inline math, should they?

@devmotion Both comments make sense to me.
1. yes we are not using cdn.mathjax.org. The point is shall we include a similar line in `_layouts/post.html` (rather than in `headers.html`)? 
2. I noticed the single and double USD thing yesterday as well. This is a bit tricky, as you can see from the html output that all inline math with $$ has actually been interpreted as a centered equation block (which is equivalent to specifying ```math ```). I think it's OK to have either single or double USD signs as inline math, the problem seems to be that MathJax is not interpreted it consistently as specified in the line `inlineMath: [['$', '$'], ['$$', '$$']]`.

PR #379, opened 2023-01-11
Use stable Julia version

### LINKED ISSUE #378: Reproducibility between Julia < 1.7 and >= 1.7
The default RNG has changed since Julia 1.7 (https://docs.julialang.org/en/v1.7-dev/NEWS/#Random) and some of notebooks with an explicit RNG (e.g. the Bayesian ODE one) are no longer reproducible as the tutorials are generated using Julia < 1.7. 
Is there a way we can code the notebook such that we set some RNG globally which works for all Julia versions?

devmotion:
This PR updates the CI setup and the Manifest files of the tutorials to the latest stable Julia version (currently Julia 1.8.5). With the recent move away from CompatHelper to a weekly update of the Manifest files (maybe could be relaxed to a two weeks interval, to reduce maintenance burden, CI time, and size of artifacts?), it seems feasible now to (try to) keep up with the stable Julia versions.

Fixes #378.


devmotion:
Closed in favour of #394.

PR #394, opened 2023-06-23
Regenerate Manifest.toml with Julia 1.9.2

devmotion:
Update of #379, let's see how many tutorials are (not) broken.

Broken/assertions fail:

- [x] Logistic Regression: Fixed by https://github.com/TuringLang/TuringTutorials/pull/394/commits/c4242604dec1245dbcf12fdc1e28755f3cd8b68d
- [x] Linear regression: Fixed by https://github.com/TuringLang/TuringTutorials/commit/2d64593d459f2fcaf40f0b19199dccc33f8d025a
- [x] Variational inference: Fixed by https://github.com/TuringLang/TuringTutorials/commit/ef59c15c698e216bb8d07a86d09c20b55656a24e
- [x] Gaussian processes: Disabled in #375, so does not matter :stuck_out_tongue: 
- [x] Seasonal time series: Fixed by https://github.com/TuringLang/TuringTutorials/pull/394/commits/33892a8659af4d4fdf53768816c2f07e7e5ce9e5

Fixes #378.

PR #399, opened 2023-07-06
AbstractMCMC fixes, rename some files, and update manifests

willtebbutt:


willtebbutt:
I believe I need additional permissions in order to do this (I created the fork to work around not having said permissions)

willtebbutt:
@yebai this looks to be mergeable

tutorials/for-developers-abstractmcmc-turing/Manifest.toml
```diff
@@ -1,1591 +1,1652 @@
 # This file is machine-generated - editing it directly is not advised
 
-julia_version = "1.6.7"
```
devmotion, line 3: @willtebbutt This Manifest file is wrong, isn't it? Currently only Julia 1.6 is supported (see eg #394).

devmotion: It seems it was generated in https://buildkite.com/julialang/turingtutorials/builds/675#01892bfc-531f-4454-ba37-1a9e3885e679 - with Julia 1.6. Apparently 1.6 is not picky about instantiating environments with Manifest files not generated explicitly for 1.6 but 1.6 will generate (and probably use) a different Manifest file. So it didn't break anything (it seems) but to me it seems it's not the "correct" Manifest file (as generally they are not guaranteed to work with different Julia versions).

PR #403, opened 2023-07-07
External samplers docs

JaimeRZP:
Docs for Turing's [#2008](https://github.com/TuringLang/Turing.jl/pull/2008).
WIP!


tutorials/15-external-samplers/external-samplers.jmd
```diff
@@ -0,0 +1,104 @@
...
+title: Using External Sampler
+permalink: /tutorials/:name/
+mathjax: true
+weave_options:
+  error : false
+---
+
+# Using External Samplers
+
+Turing currently wraps some of the most popular sampling libraries in the Julia ecosystem.
```
torfjelde, line 11: IMO no need to make claims about popularity.
```suggestion
Turing provides several wrapped samplers from external sampling libraries, e.g. HMC smaplers from AdvancedHMC.jl.
```

tutorials/15-external-samplers/external-samplers.jmd
```diff
@@ -0,0 +1,104 @@
...
+These wrappers allow new users to seemlessly sample statistical models without leaving Turing.
+However, these wrappers often lack the latests functionalities implemented on the sampling libraries that more experienced users might want to use.
```
torfjelde, line 13:
```suggestion
However, these wrappers might not always be complete, missing some functionality from the wrapped sampling library.
```

tutorials/15-external-samplers/external-samplers.jmd
```diff
@@ -0,0 +1,104 @@
...
+Moreover, users might want to use samplers currently not wrapped within Turing.
+
+For these reasons, Turing now allows users the possibility of defining their own sampler and using it to sample a Turing model.
```
torfjelde, line 16:
```suggestion
For these reasons, Turing also makes it easy to run external samplers on Turing models without any modifications or wrapping being necessary!
```

tutorials/15-external-samplers/external-samplers.jmd
```diff
@@ -0,0 +1,104 @@
...
+Let us begin by defining a simple Turing model to showcase the use of external samplers.
+For this demonstration we will use a 10 dimensional Neal's funnel:
```
torfjelde, line 18:
```suggestion
Throughout we will use a 10 dimensional Neal's funnel as a running example::
```

tutorials/15-external-samplers/external-samplers.jmd
```diff
@@ -0,0 +1,104 @@
...
+    θ ~ Truncated(Normal(0, 3), -3, 3)
+    z ~ MvNormal(zeros(d - 1), exp(θ) * I)
+    return x ~ MvNormal(z, I)
+end
+```
+
+Let us know sample the model to generate some observations and condition our model on them:
+
+```julia
+Random.seed!(1)
```
torfjelde, line 35: If it's reproducibility we're going for, we might want to use StableRNGs instead?

tutorials/15-external-samplers/external-samplers.jmd
```diff
@@ -0,0 +1,104 @@
...
+
+d = 10
+@model function funnel()
+    θ ~ Truncated(Normal(0, 3), -3, 3)
+    z ~ MvNormal(zeros(d - 1), exp(θ) * I)
+    return x ~ MvNormal(z, I)
+end
+```
+
+Let us know sample the model to generate some observations and condition our model on them:
```
torfjelde, line 32:
```suggestion
Now we sample the model to generate some observations, which we can then subsequently condition on.
```

tutorials/15-external-samplers/external-samplers.jmd
```diff
@@ -0,0 +1,104 @@
...
+
+```julia
+Random.seed!(1)
+(; x) = rand(funnel() | (θ=0,))
+model = funnel() | (; x)
+```
+
+Users can now use any sampler algorithm they wish to sample this model as long as it is of the type `AbstractMCMC.AbstractSampler`.
```
torfjelde, line 40:
```suggestion
Users can use any sampler algorithm they wish to sample this model as long as it is of the type `AbstractMCMC.AbstractSampler`.
```

Generally speaking I don't think we want to refer to "the time of writing" (unless we're describing behavior that might be changing for some reason or a bug), since users will generally be reading this without any regard for when it was written (other than assuming that it is up to date). If things change, we should change the docs; hence referene to time seems unnecessary.

tutorials/15-external-samplers/external-samplers.jmd
```diff
@@ -0,0 +1,104 @@
...
+Before discussing how is this in done in practice, it is interesting to give a high level description of the process.
+Imagine that we created an instance of external sampler that we will call `spl` such that `typeof(spl)<:AbstractMCMC.AbstractSampler`.
+In order to avoid type ambiguity within Turing, at the moment is necessary to declare `spl` as an external sampler to Turing this is done as:
```
torfjelde, line 43: I don't think this is particularly interesting to the user :confused: 

Though I do agree it might be of interest to a small minority, so might be nice to keep in. _But_ then I'd suggest moving to later in the tutorial as a description of why we have to use`externalsampler`, etc.

tutorials/15-external-samplers/external-samplers.jmd
```diff
@@ -0,0 +1,104 @@
...
+
+```julia
+espl = externalsampler(spl)
+```
+
+where `externalsampler(s::AbstractMCMC.AbstractSampler)` is a Turing function that types our external sampler adquately.
+
+Arrived at this point sampling our model is as easy as:
```
torfjelde, line 51:
```suggestion
At this point, sampling from our model is as easy as doing
```

tutorials/15-external-samplers/external-samplers.jmd
```diff
@@ -0,0 +1,104 @@
...
+
+```julia
+chain = sample(model, espl, 1_000)
+```
+
+This process will use `AbstractMCMC` functions overloaded by Turing to generate chains identical to those generated by the native samplers.
+Note that if the external sampler produces transitions that Turing cannot parse the bundling of the samples will be different or fail.
```
torfjelde, line 58: IMO this would do well with further elaboration; might even warrant it's own section, preferably with a simple example, as this is likely to be a source of confusion for the user.

tutorials/15-external-samplers/external-samplers.jmd
```diff
@@ -0,0 +1,104 @@
...
+Thankfully, we can use Turing's support for external samplers to define a HMC sampler with a custom mass matrix in AdvancedHMC and then use it to sample our Turing model.
+
+## Micro-Canonical HMC
+
+So far we have used to Turing's support for external samplers to go beyond the capabilities of the wrappers.
+Now we want to use this support to employ a sampler not supported within Turing's ecosystem yet.
+We will use the recently developped Micro-Cannoncial Hamiltonian Monte Carlo (MCHMC) sampler to showcase this.
+MCHMC[^1,^2] is HMC sampler that conserves energy along the exploration of the phase space.
+This is achieved by simulating the dynamics of a microcanonical Hamiltonian with an additional noise term to ensure ergodicity.
+The algorithm has been shown to outperform traditioanl HMC in many scenarios and it doesn't require of Metropolis adjustments.
```
torfjelde, line 99: IMO this is a bit too salespitchy :sweat_smile: 

> The algorithm has been shown to outperform traditioanl HMC

I'd prefer to not make statements like this unless we're also significantly backing these claims up. People often look to PPL docs for what's considered state-of-the-art, and so if they see this, they might think this is an endorsement of micro-canonical HMC, which I'm not quite comfortable with _at this moment_  :wink: 

> it doesn't require of Metropolis adjustments.

This is stated as a feature, but there are tons of methods (also already present in Turing, e.g. Langevin)  which does not require this, _but_ they (almost all) come at the cost of not being asymptotically (in the number of iterations) unbiased; IIUC this is also the case for micro-canonical HMC.

tutorials/15-external-samplers/external-samplers.jmd
```diff
@@ -0,0 +1,104 @@
...
+
+# Refences
+
+[^1]: Robnik et al, (Microcanonical Hamiltonian Monte Carlo)[https://arxiv.org/abs/2212.08549], 2022
+[^2]: Robnik and Seljak, (Langevine Hamiltonian Monte Carlo)[https://arxiv.org/abs/2303.18221], 2023
```
torfjelde, line 104: I'd also say, if we're adding references, then maybe worth adding some of the other methods too :shrug:

torfjelde (changes requested):
Neat:) I might add some more comments later, but I'll leave this here for now.

tutorials/15-external-samplers/external-samplers.jmd
JaimeRZP: I agree!

tutorials/15-external-samplers/external-samplers.jmd
JaimeRZP: Fair enough! I will make it more decaff.

tutorials/15-external-samplers/external-samplers.jmd
JaimeRZP: Is this a library?

tutorials/15-external-samplers/external-samplers.jmd
```diff
@@ -0,0 +1,168 @@
...
+  - There must be two `step` functions:
+    
+      + A function that performs the first step which and initializes the sampler.
+      + A function that performs the following steps and which takes and extra input, `state`, which carries the initialization information.
+
+  - The functions must follow the displayed signatures.
+  - The output of the functions must be a transition, the current state of the sampler, and a sample, what is saved to the MCMC chain.
+
+The last requirement is that the transition must be structure with a field named `θ` which constains the values of the parameters of the model for said transition.
+This allows `Turing` to seemlessly extract the parameter values at each step of the chain when bundling the chains.
```
torfjelde, line 158: This part is still something I'm not too keen on. Basically, the current way of converting a transition to something that bundle_samples can handle is really not great and I don't expect anyone outside of Turing.jl to implement this :confused:

torfjelde (changes requested):

tutorials/15-external-samplers/external-samplers.jmd
torfjelde: Yep:)

tutorials/15-external-samplers/external-samplers.jmd
```diff
@@ -0,0 +1,168 @@
...
+# Sample
+chain = sample(model, externalsampler(nuts), 10_000)
+```
+
+## Using new inference methods
+
+So far we have used to Turing's support for external samplers to go beyond the capabilities of the wrappers.
+Now we want to use this support to employ a sampler not supported within Turing's ecosystem yet.
+We will use the recently developped Micro-Cannoncial Hamiltonian Monte Carlo (MCHMC) sampler to showcase this.
+MCHMC[^3,^4] ((MCHMC's GitHub)[https://github.com/JaimeRZP/MicroCanonicalHMC.jl]) is HMC sampler that conserves energy along the exploration of the phase space.
```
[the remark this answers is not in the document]
JaimeRZP: Not extactly. hopefully this plot adds clarity.
image
I have rephrased the sentences.

tutorials/15-external-samplers/external-samplers.jmd
```diff
@@ -0,0 +1,168 @@
...
+# Sample
+chain = sample(model, externalsampler(mchmc), 10_000)
+```
+
+## The `AbstractMCMC` API
+
+As previously stated, in order to use external sampling libraries within `Turing` they must follow the `AbstractMCMC` API.
+In this section we will briefly dwell in what this entails.
+First and foremost, the inference method must be of the type `typeof(spl)<:AbstractMCMC.AbstractSampler`.
+Second, the stepping function of the MCMC algorithm must be made into instances of `AbstractMCMC.step` and follow the structure below:
```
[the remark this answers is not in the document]
JaimeRZP: Rephrased!

tutorials/15-external-samplers/external-samplers.jmd
```diff
@@ -0,0 +1,168 @@
...
+      + A function that performs the first step which and initializes the sampler.
+      + A function that performs the following steps and which takes and extra input, `state`, which carries the initialization information.
+
+  - The functions must follow the displayed signatures.
+  - The output of the functions must be a transition, the current state of the sampler, and a sample, what is saved to the MCMC chain.
+
+The last requirement is that the transition must be structure with a field named `θ` which constains the values of the parameters of the model for said transition.
+This allows `Turing` to seemlessly extract the parameter values at each step of the chain when bundling the chains.
+
+For a practical example of how to adapt a sampling library to the `AbstractMCMC` interface the readers can consult the following (code)[https://github.com/JaimeRZP/MicroCanonicalHMC.jl/blob/master/src/abstractmcmc.jl] within `MicroCanonicalHMC`.
```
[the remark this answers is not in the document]
JaimeRZP: I turned this into a list of examples.

tutorials/15-external-samplers/external-samplers.jmd
JaimeRZP: I think for the time being it is quite an important consideration for devs who want to make the connection as seemless as possible. 
We can change it as we change the code.

tutorials/15-external-samplers/external-samplers.jmd
JaimeRZP: How would you convey to the user the current need to wrap external samplers?
I am not married to the current text but I think that should be mentioned.

tutorials/15-external-samplers/external-samplers.jmd
JaimeRZP: I just removed the whole seed since I am not particularly concerned with the exact numbers.

tutorials/15-external-samplers/external-samplers.jmd
JaimeRZP: I am afraid I don't know enough about the bundling process to know what's important information to transmit to the reader. 
I also think that we can remove this sentence since user level readers will not interact with this and devs only really need to know how to define their transition object in their libraries which is covered later. 
However, please feel free to fill in information you think it might be useful.

tutorials/15-external-samplers/external-samplers.jmd
torfjelde: For certain, but it's an aspect of the implementation that is bound to change very rapidly from the current state, to the point where I really don't want to have downstream packages rely on it (the result is that we might end up making non-breaking release, which subsequently ends up breaking downstream packages because we encouraged usage of internal and volatile code :confused: )

So I'd rather just say "Right now, it's still non-trivial to make it so the resulting samples are also bundled nicely into a `MCMCChains.Chains`, but this is being worked on (see this issue)." and link to some issue where we keep track of this:)

torfjelde:
I'Ill try to have a look at this during weekend, if need be:)

JaimeRZP:
For some reason it still is using AdvancedMH@0.7.4 instead of the AdvancedMH@0.7.5

PR #414, opened 2023-07-30
Update advanced.jmd

yebai:


tutorials/docs-09-using-turing-advanced/advanced.jmd
```diff
@@ -176,6 +176,59 @@ gdemo(x) = Turing.Model(gdemo, (; x))
...
+    # Return:
+    return [x; y]
+end
+```
+
+In this case, the random variables exposed in the chain (`x_raw`, `y_raw`) are not in a helpful form — what we're after is the deterministically transformed variables `x, y`.
+
+More generally, there are often quantities in our models that we might be interested in viewing, but which are not explicitly present in our chain.
+
+We can generate draws from these variables — in this case, `x, y` — by adding them as a return statement to the model, and then calling `generated_quantities(model, chain)`. Calling this function outputs an array of values specified in the return statement of the model.
```
torfjelde, line 203: Maybe add a link to the docs for `generated_quantities`?

tutorials/docs-09-using-turing-advanced/advanced.jmd
```diff
@@ -176,6 +176,59 @@ gdemo(x) = Turing.Model(gdemo, (; x))
...
+```julia; eval=false
+generated_quantities(Neal(), chain)
+```
+
+to return an array for each posterior sample containing `x1, x2, ... x9, y`.
+
+In this case, it might be useful to reorganize our output into a matrix for plotting:
+
+```julia; eval=false
+reparam_chain = reduce(hcat, generated_quantities(Neal(), chain))'
```
torfjelde, line 222: I'm personally in favour of just keeping it as
```suggestion
reparam_chain = reduce(hcat, generated_quantities(Neal(), chain))
```
since this is a) what leads to the most efficient access, and b) (because of (a)) it's more commonly used.

The drawback is that MCMChains does not follow this convention :confused: So because of this, I'm happy to accept the current version, if you prefer it :+1:

tutorials/docs-09-using-turing-advanced/advanced.jmd
```diff
@@ -176,6 +176,59 @@ gdemo(x) = Turing.Model(gdemo, (; x))
 model = gdemo([1.5, 2.0])
 ```
 
+### Reparametrization and generated_quantities
+
+Often, the most natural parameterization for a model is not the most computationally feasible. Consider the following
+(efficiently reparametrized) implementation of Neal's funnel [(Neal, 2003)](https://arxiv.org/abs/physics/0009028):
+
+```julia; eval=false
```
torfjelde, line 184: Would it be worth making `eval=true` here? Would potentially be nice to have the outputs of this example if it doesn't take too long to run:)

torfjelde (approved):
LGTM, just added a few comments:) Nothing major though, so feel free to do with them as you wish.

shravanngoswamii:
@yebai  changes of new setup or just updated advanced.jmd

PR #422, opened 2023-09-25
Rename GP -> GPLVM

willtebbutt:
The GP tutorial is probably better referred to as a GPLVM tutorial, so I've called it that (I'm planning on adding some more straightforward GP tutorials, so this also free up the GP name for that).

Does updating the name here automatically update the name in the Turing docs, or do I also need to make a PR somewhere else?


willtebbutt:
"latent variable model" -- I think it discusses this in the tutorial itself :) I would be happy to call it "Gaussian Proess Latent Variable Model" -- it's just a bit of a long name haha

torfjelde (commented):
Do we also need to change the TOC somewhere @yebai ?

willtebbutt:
> @willtebbutt I remember that some parts of this tutorial are disabled; maybe you can try to fix them?

At some point. I'd rather keep the scope of this PR to a name change though.

willtebbutt:
Okay. So @yebai @torfjelde do we merge this, then merge https://github.com/TuringLang/turinglang.github.io/pull/56 ?

torfjelde (approved):
That seems good to me!

PR #423, opened 2023-09-25
Introduction to JuliaGPs + Turing

willtebbutt:
I've written up the tutorial I made for the BSU workshop into a TuringTutorial. I _think_ it's _okay_ to go as-is, but I suspect that people will have ideas on how it could be improved, so I'd like to iterate a bit before merging.

I would appreciate the following two kinds of feedback:
1. comments on clarity,
2. suggestions for additional analyses / modification that would improve the tutorial.

### DIFF
--- a/tutorials/15-gaussian-processes/15_gaussian_processes.jmd
+++ b/tutorials/15-gaussian-processes/15_gaussian_processes.jmd
@@ -59,10 +59,7 @@ $$
 To do this, let's define our Turing.jl model:
 
 ```julia
-using AbstractGPs, LogExpFunctions, ReverseDiff, Turing
-
-Turing.setadbackend(:reversediff) # use ReverseDiff for AD
-Turing.setrdcache(true)           # cache the tape between runs
+using AbstractGPs, LogExpFunctions, Turing
 
 @model function putting_model(d, n; jitter=1e-4)
     v ~ Gamma(2, 1)
@@ -134,10 +131,17 @@ In any case, you can judge for yourself whether you think this is the most usefu
 visualisation that we can perform -- if you think there is something better to look at,
 please let us know!
 
-Moving on, we generate samples from the posterior using the default `NUTS` sampler:
+Moving on, we generate samples from the posterior using the default `NUTS` sampler.
+We'll make use of [ReverseDiff.jl](https://github.com/JuliaDiff/ReverseDiff.jl), as it has
+better performance than [ForwardDiff.jl](https://github.com/JuliaDiff/ForwardDiff.jl/) on
+this example. See Turing.jl's docs on Automatic Differentiation for more info.
 
 ```julia
-using Random
+using Random, ReverseDiff
+
+Turing.setadbackend(:reversediff)
+Turing.setrdcache(true)
+
 m_post = m | (y=df.y,)
 chn = sample(Xoshiro(123456), m_post, NUTS(), 1_000)
 ```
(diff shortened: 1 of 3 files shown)


tutorials/15-gaussian-processes/15_gaussian_processes.jmd
```diff
@@ -0,0 +1,161 @@
...
+In this case, the only prior knowledge I have is that the proportion of successful shots
+ought to decrease monotonically as the distance from the hole increases, which should show
+up in the data as the blue lines generally going down as we move from left to right on each
+graph.
+Unfortunately, there is not a simple way to enforce monotonicity in the samples from a GP,
+and we can see this in some of the plots above, so we must hope that we have enough data to
+ensure that this relationship approximately holds under the posterior.
+In any case, you can judge for yourself whether you think this is the most useful
+visualisation that we can perform -- if you think there is something better to look at,
+please let us know!
```
torfjelde, line 135: Maybe this is crazy (and I'm def not suggesting we put this in the tutorial), but could we perform a quick linear regression and use this as a prior for the GP?

tutorials/15-gaussian-processes/15_gaussian_processes.jmd
```diff
@@ -0,0 +1,161 @@
...
+g(x) := \frac{1}{1 + e^{-x}}
+$$
+
+To do this, let's define our Turing.jl model:
+
+```julia
+using AbstractGPs, LogExpFunctions, ReverseDiff, Turing
+
+Turing.setadbackend(:reversediff) # use ReverseDiff for AD
+Turing.setrdcache(true)           # cache the tape between runs
```
torfjelde, line 65: Maybe move this to the block for the `sample` call and make a brief mention that "We'll be using `ReverseDiff` here due to better performance on this example than the default `ForwardDiff`. See [insert link to docs] for more on performance in Turing.jl." or someting?

torfjelde (commented):
LGTM!

I added two minor comments, but dope stuff:)

tutorials/15-gaussian-processes/15_gaussian_processes.jmd
willtebbutt: Hmmmm I'm not entirely sure that I understand. If I've understood correctly, the suggestion is to look at the data in order to inform our prior, which I'm not sure makes sense to me.

tutorials/15-gaussian-processes/15_gaussian_processes.jmd
willtebbutt: Great idea. Will do now

tutorials/15-gaussian-processes/15_gaussian_processes.jmd
torfjelde: So this would indeed make it, I dunno, "semi-Bayesian" or something, but it could be a useful trick to get monotonicity (though it wouldn't guarantee it). But I guess the proper way would be to just treat it as a multivariate normal and model the mean directly?

Another thing, is it sensible to put a prior on the GP prior?

torfjelde (approved):
Awesome stuff @willtebbutt :)

tutorials/15-gaussian-processes/15_gaussian_processes.jmd
willtebbutt: Well in some ways we are here -- we're placing a prior over the kernel parameters, which is implicitely placing a prior over the GP prior. Are you imagining something non-parametric though?

PR #426, opened 2023-10-18
Update 09_variational-inference.jmd

mrchaos:
`remove_names = filter(x -> !in(x, ["MPG"]), names(data))`


torfjelde (approved):
Oh that's a good catch! Thank you @mrchaos !

PR #449, opened 2024-05-24
Fixes to Bayes SDE notebook

gjgress:
Updated retcode success check to the current version used by SciMLBase. Introduced new noisy observations that are better suited for the problem, and corrected the model to calculate the likelihood based on multiple trajectories rather than a single trajectory.
#446 
https://github.com/TuringLang/Turing.jl/issues/2216

### DIFF
--- a/tutorials/10-bayesian-stochastic-differential-equations/index.qmd
+++ b/tutorials/10-bayesian-stochastic-differential-equations/index.qmd
@@ -17,6 +17,7 @@ Pkg.instantiate();
 ```{julia}
 using Turing
 using DifferentialEquations
+using DifferentialEquations.EnsembleAnalysis
 
 # Load StatsPlots for visualizations and diagnostics.
 using StatsPlots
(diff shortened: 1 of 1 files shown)


THargreaves:
It's my understanding that pseudo-marginal HMC, as introduced in Alenlöv et al., 2016, is a bit more nuanced than what is happening here. Even if you have an unbiased estimate of the likelihood, that doesn't mean your gradient estimates are unbiased and so your dynamics are incorrect.

The approach taken in the PM-HMC paper is quite a bit more complicated. To avoid having a confusing/misleading example in the docs, it might be wiser to fit this model using MH, as in the PM-MH framework which is a valid "exact approximate" algorithm.

That said, you could stick with the current approach but I think you'd need to make it clear that this an approximate method.

gjgress:
The discussion on pseudo-marginal samplers or sequential Monte Carlo methods is not something I'm qualified to comment on, nor implement myself. From the viewpoint of creating a tutorial, I would say that using either of those tools is outside the scope of this tutorial, and ought to be its own tutorial (if its even within the scope of Turing as a whole-- I could not find an implementation of either of these methods using Turing during my search). 

Of course if it can be done it would be fantastic-- based on what others have said, it seems that except under particular conditions, it is generally bad form to do the method here, and so using this method without at least mentioning its approximate nature may be misleading to less experienced users trying to use Turing for their SDE problems.

In any case, there isn't more for me to add to this tutorial, as I am not knowledgeable enough on this to add that commentary to the tutorial.

PR #468, opened 2024-05-31
Alternative tutorial on implementing samplers

torfjelde:
I did a talk on Turing.jl for a statistical audience earlier today and described how it's possible to implement a samplers in AbstratMCMC.jl in such a way that it's very easy to use the resulting sampler to target both Turing.jl and Stan models (through BridgeStan).

Afterwards I was asked if there was some tutorial or something online outlining how to do this in a detailed way. After having a look at the current docs on how to use the `externalsampler`, I felt they were a bit overly complicated and didn't _quite_ get all the points across, and so I did a rewrite of it.

I think this version will be very, very useful for inference researchers who want their methods to become available for a larger userbase:)

EDIT: Note that I haven't added example with StanLogDensityProblems.jl, but I think we maybe should, as it would be quite nice.

I'm also thinking we should make the example a bit more complicated later in the tutorial, e.g. adding stuff like "how to test the sampler with MCMCTesting.jl" from @Red-Portal , adding a few more things to the `MALAState`, e.g. `isaccept` to indicate whether it got accepted, maybe add some adaptation, etc. But I think the current version should be merged first, and then we can build on it later :+1:


torfjelde:
But the current docs are somewhat out of date + it's too similar in what they describe IMO. Maybe it makes more sense to add another section at the end of this tutorial demonstrating some examples, where I'll add back the examples in the original?

torfjelde:
> Maybe it makes more sense to add another section at the end of this tutorial demonstrating some examples, where I'll add back the examples in the original?

Happy with this Hong?

torfjelde:
Done :+1:

torfjelde:
@yebai notice that I've implemented a version of MALA which uses a single leapfrog step rather than the "standard" formulation of MALA. My original intention was to add a final section on a more "complex" version being autoMALA. Buuut I'm uncertain if this is worth it or not. Should I still pursue this or just leave it for now?

tutorials/docs-17-implementing-samplers/index.qmd
```diff
@@ -0,0 +1,479 @@
...
+
+Luckily for us, there is a package called [LogDensityProblems.jl](https://github.com/tpapp/LogDensityProblems.jl) which provides an interface for *exactly* this!
+
+To demonstrate how one can implement the "[LogDensityProblems.jl](https://github.com/tpapp/LogDensityProblems.jl) interface"[^1] we will use a simple Gaussian model as an example:
+
+```{julia}
+using LogDensityProblems: LogDensityProblems;
+
+# Let's define some type that represents the model.
+struct NormalLogDensity{M<:AbstractVector{<:Real}}
```
mhauru, line 84: Might it make sense to call this something else, like `IsotropicNormalModel` or something? It took me a moment to understand that this represents just the Gaussian model for which we can compute log densities, not a log density calculation.

tutorials/docs-17-implementing-samplers/index.qmd
```diff
@@ -0,0 +1,479 @@
...
+
+struct MALA{T,A} <: AbstractMCMC.AbstractSampler
+    "stepsize used in the leapfrog step"
+    ϵ_init::T
+    "covariance matrix used for the momentum"
+    M_init::A
+end
+```
+
+Notice how we've added the suffix `_init` to both the stepsize and the covariance matrix. We've done this because a `AbstractMCMC.AbstractSampler` should be *immutable*. Of course there might be many scenarios where we want to allow something like the stepsize and / or the covariance matrix to vary between iterations, e.g. during the burn-in / adaptation phase of the sampling process we might want to adjust the parameters using statistics computed from these initial iterations. But information which can change between iterations <span class="underline">should not go in the sampler itself</span>! Instead, this information should go in the sampler *state*.
```
mhauru, line 165: Might the underline be better replaced with `*`? I tried to hover over it thinking it might be a link.

tutorials/docs-17-implementing-samplers/index.qmd
```diff
@@ -0,0 +1,479 @@
...
+2.  A state `MALAState` for our sampler `MALA`.
+
+That means that we're ready to implement the only thing that really matters: `AbstractMCMC.step`.
+
+`AbstractMCMC.step` defines the MCMC iteration of our `MALA` given the current `MALAState`. Specifically, the signature of the function is as follows:
+
+```{julia}
+#| eval: false
+function AbstractMCMC.step(
+    # The RNG to ensure reprodicibility.
```
mhauru, line 194:
```suggestion
    # The RNG to ensure reproducibility.
```

tutorials/docs-17-implementing-samplers/index.qmd
```diff
@@ -0,0 +1,479 @@
...
+    # Additional keyword arguments that we may or may not need.
+    kwargs...
+)
+```
+
+Moreover, there is a specific `AbstractMCMC.AbstractModel` which is used to indicate that the model that is provided implements the [LogDensityProblems.jl](https://github.com/tpapp/LogDensityProblems.jl) interface: `AbstractMCMC.LogDensityModel`.
+
+Since, as we discussed earlier, in our case we're indeed going to work with types that support the [LogDensityProblems.jl](https://github.com/tpapp/LogDensityProblems.jl) interface, we'll define `AbstractMCMC.step` for such a `AbstractMCMC.LogDensityModel`.
+
+Note that `AbstractMCMC.LogDensityModel` has no other purpose; it has a single field called `logdensity`, and it does nothing else. But by wrapping the model in `AbstractMCMC.LogDensityModel`, it allows samplers which wants to work with [LogDensityProblems.jl](https://github.com/tpapp/LogDensityProblems.jl) to define their `AbstractMCMC.step` on this type without running into method ambiguities.
```
mhauru, line 211:
```suggestion
Note that `AbstractMCMC.LogDensityModel` has no other purpose; it has a single field called `logdensity`, and it does nothing else. But by wrapping the model in `AbstractMCMC.LogDensityModel`, it allows samplers that want to work with [LogDensityProblems.jl](https://github.com/tpapp/LogDensityProblems.jl) to define their `AbstractMCMC.step` on this type without running into method ambiguities.
```

tutorials/docs-17-implementing-samplers/index.qmd
```diff
@@ -0,0 +1,479 @@
...
+    ∇logγ_x = last(LogDensityProblems.logdensity_and_gradient(model, x))
+    p1 = p + (ϵ / 2) .* ∇logγ_x
+    # Update the "position" `x` using momentum `p1`.
+    x̃ = x + ϵ .* (M \ p1)
+    # Update momentum `p1` using position `x̃`
+    ∇logγ_x̃ = last(LogDensityProblems.logdensity_and_gradient(model, x̃))
+    p2 = p + (ϵ / 2) .* ∇logγ_x̃
+    # Flip momentum `p2`.
+    p̃ = -p2
+    # Return.
```
mhauru, line 295:
```suggestion
```
The return statement seems self-documenting.

tutorials/docs-17-implementing-samplers/index.qmd
```diff
@@ -0,0 +1,479 @@
...
+```{julia}
+function leapfrog_step(model, x, p, ϵ, M)
+    # Update momentum `p` using "position" `x`.
+    ∇logγ_x = last(LogDensityProblems.logdensity_and_gradient(model, x))
+    p1 = p + (ϵ / 2) .* ∇logγ_x
+    # Update the "position" `x` using momentum `p1`.
+    x̃ = x + ϵ .* (M \ p1)
+    # Update momentum `p1` using position `x̃`
+    ∇logγ_x̃ = last(LogDensityProblems.logdensity_and_gradient(model, x̃))
+    p2 = p + (ϵ / 2) .* ∇logγ_x̃
```
mhauru, line 292: Should this be 
```suggestion
    p2 = p1 + (ϵ / 2) .* ∇logγ_x̃
```
? I think that's what it is in the equations at the start of the page. I know nothing about these sampling algorithms though, so correct me if I'm wrong.

tutorials/docs-17-implementing-samplers/index.qmd
```diff
@@ -0,0 +1,479 @@
...
+samples_array = stack(map(stack, samples))
+```
+
+But the fact that we have to provide the `AbstractMCMC.sample` call, etc. with an `initial_state` to get started is a bit annoying. We can avoid this by also defining a `AbstractMCMC.step` *without* the `state` argument:
+
+```{julia}
+function AbstractMCMC.step(
+    rng::Random.AbstractRNG,
+    model_wrapper::AbstractMCMC.LogDensityModel,
+    sampler::MALA;
```
mhauru, line 366:
```suggestion
    ::MALA;
```

tutorials/docs-17-implementing-samplers/index.qmd
```diff
@@ -0,0 +1,479 @@
...
+    # NOTE: No state provided!
+    kwargs...
+)
+    model = model_wrapper.logdensity
+    # Let's just create the initial state by sampling using  a Gaussian.
+    x = randn(rng, LogDensityProblems.dimension(model))
```
mhauru, line 372: This is only valid for our particular model, right? In that this may e.g. give negative values, which might not be valid parameters for some models. Don't want to overcomplicate the example, but might it be worth mentioning that in general one should e.g. sample from the prior?

EDIT: Noticed that you address this later. Maybe when you address it mention that the fix would be to change this method of `step`?

tutorials/docs-17-implementing-samplers/index.qmd
```diff
@@ -0,0 +1,479 @@
...
+As we promised, all of this hassle of implementing our `MALA` sampler in a way that uses [LogDensityProblems.jl](https://github.com/tpapp/LogDensityProblems.jl) and [AbstractMCMC.jl](https://github.com/TuringLang/AbstractMCMC.jl) gets us something more than *just* an "automatic" implementation of `AbstractMCMC.sample`.
+
+It also enables use with Turing.jl through the `externalsampler`, but we need to do one final thing first: we need to tell Turing.jl how to extract a vector of parameters from the "sample" returned in our implementation of `AbstractMCMC.step`. In our case, the "sample" is just a `Vector{<:Real}`, so we just need the following line:
+
+```{julia}
+# Load Turing.jl.
+using Turing
+
+# Overload the `getparams` method for our "sample" type, which is just a vector.
+Turing.Inference.getparams(::Turing.Model, x::AbstractVector{<:Real}) = x
```
mhauru, line 397: I'm a bit confused by this, which looks like type piracy to me. Should this rather be defined for `MALAState`?

EDIT: Also just noticed that the docstring of `getparams` says "Return a named tuple of parameters", which adds to my confusion.

tutorials/docs-17-implementing-samplers/index.qmd
```diff
@@ -0,0 +1,479 @@
...
+# our `MALA` sampler in the `externalsampler` to tell Turing.jl that the sampler
+# expects something that implements LogDensityProblems.jl.
+chain = sample(turing_model, externalsampler(sampler), 10_000; progress=false)
+```
+
+Pretty neat, eh?
+
+### Models with constrained parameters
+
+One thing we've sort of glossed over in all of the above is that MALA, at least how we've implemented, requires $x$ to live in $\mathbb{R}^d$ for some $d > 0$. If some of the parameters were in fact constrained, e.g. we were working with a `Beta` distribution which has support on the interval $(0, 1)$, *not* on $\mathbb{R}^d$, we could easily end up outside of the valid range $(0, 1)$.
```
mhauru, line 417:
```suggestion
One thing we've sort of glossed over in all of the above is that MALA, at least how we've implemented it, requires $x$ to live in $\mathbb{R}^d$ for some $d > 0$. If some of the parameters were in fact constrained, e.g. we were working with a `Beta` distribution which has support on the interval $(0, 1)$, *not* on $\mathbb{R}^d$, we could easily end up outside of the valid range $(0, 1)$.
```

tutorials/docs-17-implementing-samplers/index.qmd
```diff
@@ -0,0 +1,479 @@
...
+
+As expected, we run into a `DomainError` at some point, while if we set `unconstrained=true`, letting Turing.jl transform the model to a unconstrained form behind the scenes, everything works as expected:
+
+```{julia}
+sample(demo(), externalsampler(sampler; unconstrained=true), 10_000; progress=false)
+```
+
+Neat!
+
+Similarly, which automatic differentiation backend one should use can be specified through the `adtype` keyword argument too. For example, if we want to use [ReverseDiff.jl](https://github.com/JuliaDiff/ReverseDiff.jl) instead of the default [ReverseDiff.jl](https://github.com/JuliaDiff/ForwardDiff.jl):
```
mhauru, line 456:
```suggestion
Similarly, which automatic differentiation backend one should use can be specified through the `adtype` keyword argument too. For example, if we want to use [ReverseDiff.jl](https://github.com/JuliaDiff/ReverseDiff.jl) instead of the default [ForwardDiff.jl](https://github.com/JuliaDiff/ForwardDiff.jl):
```

tutorials/docs-17-implementing-samplers/index.qmd
```diff
@@ -0,0 +1,479 @@
...
+
+Double-neat.
+
+## Summary
+
+At this point it's worth maybe reminding ourselves what we did and also *why* we did it:
+
+1.  We define our models in the [LogDensityProblems.jl](https://github.com/tpapp/LogDensityProblems.jl) interface because it makes the sampler agnostic to how the underlying model is implemented.
+2.  We implement our sampler in the [AbstractMCMC.jl](https://github.com/TuringLang/AbstractMCMC.jl) interface, which just means that our sampler is a subtype of `AbstractMCMC.AbstractSampler` and we implement the MCMC transition in `AbstractMCMC.step`.
+3.  (1) and (2) makes it so our sampler can be used with a wide range of model implementations, amongst them being models implemented in both Turing.jl and Stan. This gives you, the inference implementer, a large collection of models to test your inference method on, in addition to allowing users of Turing.jl and Stan to try out your inference method with minimal effort.
```
mhauru, line 477: Quarto renders the `(1) and (2)` as "1. and (2)" with a funny indent. I think technically the `(1)` starts another nested enumerated list.

mhauru (changes requested):
Great stuff, I learned a lot from reading it. I only have some localised comments and typo fixes.

tutorials/docs-17-implementing-samplers/index.qmd
torfjelde: Good idea:)

tutorials/docs-17-implementing-samplers/index.qmd
torfjelde: Even if you use a different sample method here which samples values inside the support of the distribution, you'd still run into the same bug (just further down the callstack) because a gradient-based sampler like MALA will inevitably end up exploring the space outside of the support (unless you have a _tiny_ stepsize). So we can't really implement this sampler for models (in a generic fashion, without a lot of gluecode) in such a way that it also works with constrained distributions.

Heeeence I just avoid the issue entirely, demosntrate a case later and how we can use Turing.jl to "address" this.

torfjelde:
This should be ready now:)

PR #491, opened 2024-07-09
Fixed Flux typo in Bayesian Neural Networks doc page

Qfl3x:


Qfl3x:
Perhaps also change `parameters` to `ps` in the text as it is a bit confusing.

PR #497, opened 2024-07-28
Improvements to docs infrastructure

### LINKED ISSUE #495: Manifest.toml files are outdated
The `Manifest.toml` files on the master branch still point to 0.32.0 of Turing, which leads to e.g. this `UndefVarError`: https://turinglang.org/docs/tutorials/docs-12-using-turing-guide/#maximum-likelihood-and-maximum-a-posterior-estimates

https://github.com/TuringLang/docs/blob/1f797b78268685c853efb5a32c3c71744152e47b/tutorials/docs-12-using-turing-guide/Manifest.toml#L1882-L1886

I notice that there used to be [a weekly GitHub Action to update the manifests](https://github.com/TuringLang/docs/blob/9e4f6f455eda8d73faf795f9beb632e009e348dc/.github/workflows/UpdateManifests.yml), which was removed as part of #441. So a quick fix would be to re-enable that workflow and run it.

However, I think a more robust solution for the long term would be to:

1. Use a single environment for building all the docs, rather than one environment per notebook. I don't think there is any use case for having separate environments, and even if there _were_ I think that it would not be the right thing to do — it would be very confusing to the user if the code on separate pages could not be run in the same environment.

2. Pin the version of Turing with a `[compat]` block in the `Project.toml` file (on both the master and backport branches)

3. In the main documentation workflow (using `on: push`), add a check to make sure that the version of Turing specified in `Project.toml` matches the version specified in `_quarto.yml`

4. ~~Gitignore the manifest, only regenerating it when building the docs. IMO it makes more sense to let Julia resolve to the latest compatible versions (e.g. in order to use the latest bugfix version of Turing; and after all, a user reading the docs and running the code is, in all likelihood, going to fetch the latest versions anyway instead of looking for the manifest in this repo). Also, this spares us this issue of having to update the manifest when a new version of Turing is released.~~

Overall, this proposal means that when a new version of Turing is launched, only three files (`_quarto.yml`, `Project.toml`, and `Manifest.toml`) need to be changed; and GitHub will tell you right away if you forget to change one of them. (Of course, the list of old versions will also need to be updated, but that is a separate matter, unrelated to this issue.)

### LINKED ISSUE #496: Quarto freeze files shouldn't be loaded from cache if environment is different
In the current workflow, Quarto's `_freeze` folder is cached:

https://github.com/TuringLang/docs/blob/99a969449d1ad364c74b69e9857702463e3a19b3/.github/workflows/publish.yml#L29-L37

This is great as it avoids rerunning all the code if the source qmd files haven't changed (cf. [Quarto docs](https://quarto.org/docs/projects/code-execution.html#using-freeze)).

However, as far as I can tell, if the environment changes (i.e. the `Manifest.toml`) but the source files aren't changed, the code won't be rerun. The correct behaviour should be to rerun if either the source or the environment is changed.

This can be done by hashing the manifests and adding this to the run key: https://docs.github.com/en/actions/learn-github-actions/expressions#hashfiles and also removing the restore-key.

Generally, this means that the docs deployment is more likely to take 2 hours rather than 2 minutes. However, I think that correctness is most important here :)

#497 should fix this.

penelopeysm:
- Closes #495
- Closes #496
- Also fixes some typos in https://turinglang.org/docs/versions.html ('Documention' -> 'Documentation')
- Adds some extra files that should be gitignored (Quarto requires a python venv to run, so I added the typical `venv` and `.venv` paths)

Please see issues for more details!


penelopeysm:
Unfortunately, GitHub is still running the version of the workflows on the main branch, so the current PR preview doesn't show the correctly built docs (for example, it still has the `UndefVarError` I mentioned in #495).

To get around this, I'm going to temporarily add a new workflow file that runs on this branch, then remove it after it's done running.

penelopeysm:
(@shravanngoswamii Requesting your expertise on github actions here :))

.github/workflows/preview.yml
```diff
@@ -22,29 +22,30 @@ jobs:
         with:
           version: '1.10'
 
+      - name: Instantiate Julia environment
```
shravanngoswamii, line 25: We were instantiating environment in quarto file itself so we will do it this way then that should be removed I guess! See https://github.com/TuringLang/docs/blob/1f797b78268685c853efb5a32c3c71744152e47b/tutorials/00-introduction/index.qmd#L8C1-L13C4

.github/workflows/preview.yml
```diff
@@ -22,29 +22,30 @@ jobs:
...
+        run: julia --project=. -e 'using Pkg; Pkg.instantiate()'
```
shravanngoswamii, line 26: We were instantiating environment in quarto file itself so we will do it this way then that should be removed I guess! See https://github.com/TuringLang/docs/blob/1f797b78268685c853efb5a32c3c71744152e47b/tutorials/00-introduction/index.qmd#L8C1-L13C4

shravanngoswamii: Also, I can't see project instantiation in publish workflow!

.github/workflows/version_check.yml
shravanngoswamii, line 1: The reason we added version check was to suggest the updated version in PRs, this one looks more clean and better approach, but should we add suggestion part too in this or not? @yebai should have some opinion on it!

assets/scripts/versions.sh
shravanngoswamii, line 1: Oh no, how did the spelling of 'Documentation' sneak past us?😅 Thanks for fixing that!

.github/workflows/preview.yml
shravanngoswamii: Also, I think we did tried this approach at first, and we faced some issues back then, I don't remember the particulars but this time it looks fine to me, let's try it!

penelopeysm: Oh, yes, you're right! Still need to instantiate in the publish workflow.

If it's a single env then instantiating it one time at the start works, I tested it [once here](https://github.com/TuringLang/docs/actions/runs/10131613167/job/28014548724?pr=497) :)

.github/workflows/version_check.yml
penelopeysm: Yes, this is definitely missing that which was a really neat feature.

I have some hesitance in adding a suggested correction on the PR, because if there are two values it's non-triviail to figure out which one is supposed to be the correct one. (We could say it's always the more recent one; but this wouldn't work on the `backport-0.32` branch, for example.)

However, I'm happy to be convinced otherwise.

penelopeysm: Okay, I've done that ^ for now! It's a bit problematic with the Manifest, though, because the diffs for that file are very intense (and it gets confused) :(

https://github.com/TuringLang/docs/actions/runs/10361000353/job/28680473684

penelopeysm: Since the diffs for the Manifest file are too large to handle via review suggestions, I've made the workflow instead suggest changes as a new PR. It also drops a comment on the PR to notify the author. I think that should handle our most common use case for this workflow.

shravanngoswamii:
@penelopeysm after this is merged, what will be the ideal process of updating the documentation of old versions!

penelopeysm:
@shravanngoswamii That's a good point. Am I right in saying that only 0.32 is generated via this repo?

If that's the case, I suspect the only change we need to make on the `backport-v0.32` branch is to coalesce the envs into a single Project.toml with the appropriate compat entry for Turing, and it should work fine with the present workflow.... I think.

shravanngoswamii:
Not only v0.32, let's say we are on v0.36 and some changes are required in v0.34 (which is not backport)!

penelopeysm:
Hm. I had kind of assumed that after we upgrade master to 0.34, we would then also create a backport-v0.33 branch that would just be the contents of master prior to upgrade. Those could then be edited however we wish and should be published just fine by the current workflow.

shravanngoswamii:
> I had kind of assumed that after we upgrade master to 0.34, we would then also create a backport-v0.33

Actually, we created the backport of v0.32 because it was released when I was developing versions workflows for this repo!
See here, we are publishing rendered docs of old versions in gh-pages itself:
https://github.com/TuringLang/docs/blob/b65e747ccc8ea125b83a1db6f3de80c0bb839947/.github/workflows/publish.yml#L45C1-L52C55

https://github.com/TuringLang/docs/blob/b65e747ccc8ea125b83a1db6f3de80c0bb839947/.github/workflows/publish.yml#L93C1-L99C23


gh-pages: https://github.com/TuringLang/docs/tree/gh-pages/versions

shravanngoswamii:
In current workflow, we have to create a branch from some old commit and then update the old docs by changing version number in _quarto.yml

penelopeysm:
> In current workflow, we have to create a branch from some old commit and then update the old docs by changing version number in _quarto.yml

Yeah. I think we are agreeing, right? ;) The docs for v0.33 would live in a v0.33 branch (we can call it backport-v0.33 for consistency, or we can take a different name, not a big deal). The `Project.toml` and `_quarto.yml` on that branch would be kept at 0.33. Then we make edits on that branch if we want to change old docs. As long as we keep the structure of those branches in agreement with the publish workflow, I don't see any problems arising.

shravanngoswamii:
Is it fine to create a lot of branches, because they will keep increasing!

penelopeysm:
Personally I wouldn't have a problem with it! As long as we are publishing multiple versions of the docs online, the source for those versions will have to live _somewhere_, and this repo is the most logical place to house them.

Most software projects I know do this too, for example the Julia core repo has [release/vX.Y branches](https://github.com/JuliaLang/julia/branches/all?query=release) with all the docs which (I presume) is how they keep docs for old versions online :)

If we find one day that there are too many branches, then we could just stop providing online documentation for old versions. For example, we could say that we provide docs only for the 5 latest versions. We could discuss that in the future, but I can't say I feel particularly bad about using GitHub's disk space haha.

shravanngoswamii:
> We could discuss that in the future, but I can't say I feel particularly bad about using GitHub's disk space haha.

Well, I agree to all your points! I think this PR is good to go!!

shravanngoswamii (approved):

shravanngoswamii:
@penelopeysm Thank you for all the effort!!

.github/workflows/version_check.yml
```diff
@@ -0,0 +1,198 @@
...
+          import YAML
+          import TOML
+          import JSON
+          import HTTP
+
+          PROJECT_TOML_PATH = "Project.toml"
+          QUARTO_YML_PATH = "_quarto.yml"
+          MANIFEST_TOML_PATH = "Manifest.toml"
+
+          function major_minor_match(vs...)
```
mhauru, line 58: Is this equivalent to `all(v[1] == v for v in vs)` except it ignores `patch`, `prerelease`, and `build`? Could it have a short docstring to say as much?

.github/workflows/version_check.yml
```diff
@@ -0,0 +1,198 @@
...
+          quarto_version = VersionNumber(quarto_version_str)
+          println("_quarto.yml version: ", quarto_version_str)
+
+          project_toml = TOML.parsefile(PROJECT_TOML_PATH)
+          project_version_str = project_toml["compat"]["Turing"]
+          project_version = VersionNumber(project_version_str)
+          println("Project.toml version: ", project_version_str)
+
+          manifest_toml = TOML.parsefile(MANIFEST_TOML_PATH)
+          manifest_version = VersionNumber(manifest_toml["deps"]["Turing"][1]["version"])
```
mhauru, line 118: What does the `[1]` do? Does it presume some fixed ordering of things in the Manifest?

.github/workflows/version_check.yml
```diff
@@ -0,0 +1,198 @@
+name: Check Turing.jl version consistency
```
mhauru, line 1: Could this file have a more extensive comment at the top, explaining what the action does?

.github/workflows/version_check.yml
```diff
@@ -0,0 +1,198 @@
...
+
+              if !major_minor_match(latest_version, project_version)
+                  push!(errors, "$(PROJECT_TOML_PATH) out of date")
+                  println("$(PROJECT_TOML_PATH) is out of date; updating")
+                  update_project_toml(PROJECT_TOML_PATH, latest_version)
+              end
+
+              if !major_minor_match(latest_version, quarto_version)
+                  push!(errors, "$(QUARTO_YML_PATH) out of date")
+                  n_errors += 1
```
mhauru, line 139: This doesn't seem to be defined or used.

.github/workflows/version_check.yml
```diff
@@ -0,0 +1,198 @@
...
+              if isempty(errors)
+                  println("All good")
+              else
+                  error("The following errors occurred during version checking: \n", join(errors, "\n"))
+              end
+
+          else
+              # Don't attempt to fetch the latest version; just check for consistency
+              # and error if versions don't match.
+              if !major_minor_match(quarto_version, project_version, manifest_version)
```
mhauru, line 169: Is there a significance to this ignoring `patch` for `manifest_version`, but the earlier if-branch not doing so?

mhauru (changes requested):
Had a few questions and small comments, seems great. That auto-version-updater GHA is faaancy! I wonder if the Julia code for it could live in a `.jl` file somewhere. Having that much Julia code as a string in yaml feels funny, and exempts it from things like linting.

.github/workflows/version_check.yml
penelopeysm: Oops.

.github/workflows/version_check.yml
penelopeysm: Yes, basically, _quarto.yml and Project.toml don't include patch numbers, but the Manifest does. So there's no source of truth for comparison.

On the other branch, we can get the desired patch number from the GitHub API. In principle we could get it here too (by enumerating every tag on GitHub, filtering for the ones with the correct major+minor, then picking the one with the largest patch number). It's a bit of work because to do it correctly we have to handle the result pagination from GitHub's API response, so I didn't bother doing that. Should we?

.github/workflows/version_check.yml
penelopeysm: I'm not fluent in TOML, but I think it's because every entry in the Manifest is specified like

```toml
[[deps.Turing]]
version = "whatever"
```

and the `[1]` comes from having to unpack the extra level of nesting with the double-square-brackets.

penelopeysm: I wonder if there's any case where this might not happen. Maybe if there are two packages named Turing but with different uuids?!

penelopeysm: Pkg refuses to add two packages with the same name (I just tried). No idea what purpose the extra level of nesting serves, then; but given that I failed to come up with an example pathological enough to break it, I no longer feel too bad about this magic `[1]`.

mhauru: Even if there is some case I'm not aware of when you might have duplicate entries, I don't think it's going to happen to the dependency called `Turing` in this repo. Happy with this.

.github/workflows/version_check.yml
mhauru: Naah, I think that's fine. Thanks for the explanation.

.github/workflows/version_check.yml
penelopeysm: I've changed the implementation to use all(). Do you think it still needs a docstring to explain what the code does?

mhauru (approved):
Looks great, thanks Penny!

penelopeysm:
Thanks everyone!

PR #504, opened 2024-08-13
Fix typo from "Flux" to "Lux"

Dale-Black:


penelopeysm (approved):
@Dale-Black Thank you! :)

PR #513, opened 2024-08-30
Update Turing to v0.34

penelopeysm:
This PR follows on from the release of v0.34.0: https://github.com/TuringLang/Turing.jl/releases/tag/v0.34.0

- It also includes some small textual improvements to the Contributing page
- I've also made a new branch, `backport-v0.33`, with the current contents of master, so that the v0.33 docs get published as well.

PR #514, opened 2024-08-30
Reorganise introductory material

penelopeysm:
Don't merge this, do #513 first, then reset the target branch to master :)


tutorials/docs-00-getting-started/index.qmd
```diff
@@ -16,96 +16,149 @@ Pkg.instantiate();
...
 ```
+
+and the distribution of the variance:
+
+```{julia}
+density([s[2] for s in analytical_samples]; label="Posterior (Analytical)")
+density!(chn[:s²]; label="Posterior (HMC)")
+```
+
+we see that our HMC sampler has given us a good representation of the posterior distribution.
```
penelopeysm, line 164: In fact, both the HMC sample as well as our custom sampler are affected by randomness and any time `quarto render` is run, there's no guarantee as to how well the distributions will line up.

This would of course be improved with more samples, but then the docs take longer to run.

Would it make sense to set the RNG seed where necessary?

penelopeysm (commented):
Explanation of what I did + one question at the end.

penelopeysm:
Will close and reopen this PR later, it's too messy!

PR #522, opened 2024-09-14
Split up 'advanced usage' page

penelopeysm:
This PR splits up the [Advanced Usage page](https://turinglang.org/docs/tutorials/docs-09-using-turing-advanced/), making each section a page on its own.

I'm proposing this for a couple of reasons:

1. It's easier for people to find what they are looking for via the sidebar.
2. Having one section for "miscellany" tends to lead to clutter. Right now it's not very long but it could in principle keep growing and we probably don't want that.

Some implementation details of this PR which may warrant discussion:

1. I'm aware we shouldn't try to break existing links. Therefore, I've kept the `docs-09-using-turing-advanced` notebook (Quarto will still build it so the link will still work, see e.g. [the PR preview docs](https://turinglang.org/docs/pr-previews/522/tutorials/docs-09-using-turing-advanced/)), but excluded it from the Quarto ToC. The contents of that page have been replaced with links to the new pages.

2. I wasn't entirely sure what to do with the [Task Copying section](https://turinglang.org/docs/tutorials/docs-09-using-turing-advanced/#task-copying). Right now I've left it out, because it doesn't seem to be a piece of info that would make a difference to an end user. If there's something about this that impacts the user, it'd be great if we could expand upon it with examples.

3. I chose new names for the folders that don't really obey the existing structure.

   ```
   usage-custom-distribution/
   usage-generated-quantities/
   usage-modifying-logprob/
   dev-model-manual/
   ```

   whereas most of the docs are `docs-XY-...`. I think the numbers have kind of outgrown their usefulness, some of them have been shuffled around in the ToC and generally they aren't very future proof. The only benefit of sticking to the original naming structure is, imo, it looks prettier when you type `ls` ;)


tutorials/usage-generated-quantities/index.qmd
```diff
@@ -0,0 +1,67 @@
...
+
+For example, in the above reparametrization, we sample from our model:
+
+```{julia}
+setprogress!(false)
+chain = sample(Neal(), NUTS(), 1000)
+```
+
+Notice that only `x_raw` and `y_raw` are stored in the chain; `x` and `y` are not because they do not appear on the left-hand side of a tilde-statement.
+
```
penelopeysm, line 47: I added this line. Would appreciate if somebody can confirm that this is an accurate statement!

tutorials/usage-modifying-logprob/index.qmd
```diff
@@ -0,0 +1,57 @@
...
+using Turing
+using LinearAlgebra
+
+@model function demo(x)
+    m ~ MvNormal(zero(x), I)
+    if dot(m, x) < 0
+        Turing.@addlogprob! -Inf
+        # Exit the model evaluation early
+        return nothing
+    end
```
penelopeysm, line 41: Also, by 'rejecting samples', can I check that this really means 'forcing the sampler to reject the samples'? In that we're relying on the implementation of the sampler doing this, rather than (e.g.) some AbstractMCMC code that checks the log probability and discards the sample if it's `-Inf`?

penelopeysm (commented):
Also some minor questions about the contents of the docs.

tutorials/usage-modifying-logprob/index.qmd
torfjelde: That is indeed correct:)

tutorials/usage-generated-quantities/index.qmd
torfjelde: This is indeed an accurate statement:)

torfjelde (approved):
All looks good to me @penelopeysm ; good stuff!

One thing on the generated quantities section: we now also have `:=` which also provides an alterantive way of keeping track of non-tilde values. It might be worth adding a section on this (though in a separate PR) :)

tutorials/usage-modifying-logprob/index.qmd
penelopeysm: Gotcha! I'll add a few words to that effect, then merge.

PR #529, opened 2024-10-03
Remove `error: true` from config

### LINKED ISSUE #525: Quarto doesn't complain if examples error
Right now Quarto doesn't complain / exit with an error if any of the code examples return an error. which leads to things like https://turinglang.org/docs/tutorials/03-bayesian-neural-network/. Probably related to the use of the new Julia engine

It's not immediately obvious how to fix this, because maybe sometimes we _do_ want the error.

I can look into it.

penelopeysm:
Closes #525


.github/workflows/preview.yml
```diff
@@ -25,6 +25,9 @@ jobs:
         with:
           version: '1.10'
 
+      - name: Load Julia packages from cache
+        uses: julia-actions/cache@v2
+
       # Note: needs resolve() to fix #518
```
penelopeysm, line 31: Once the Quarto `_freeze` cache has been generated for a particular Manifest, the main time it takes to build the docs comes from project instantiation (~ 10 mins) I don't know if this would help it, but I feel like it's worth a shot.

shravanngoswamii (approved):
Thank you, @penelopeysm!

PR #537, opened 2024-10-11
Add restore key to _freeze cache key

### LINKED ISSUE #536: Cache _freeze folder more intelligently
The `_freeze` folder is cached in our current CI runs, but the cache isn't ever updated (because GitHub Actions is weird), meaning that only the first run with a given Manifest file is cached. This means that any time a page is changed it will be rerun on every subsequent CI run, which can take a long time.

julia-actions/cache gets around this by using the GitHub API to delete a cache before trying to save it. We should probably try to do something similar.

https://github.com/julia-actions/cache/blob/824243901fb567ccb490b0d0e2483ccecde46834/action.yml#L173-L184

penelopeysm:
This PR makes a tiny change to both preview and publish workflows:

https://github.com/TuringLang/docs/blob/d7dd920038893dee8fc9f29d552f397ec53ececd/.github/workflows/publish.yml#L40-L49

The main cache key itself will now be determined by both the Manifest as well as the contents of the notebooks (`**/index.qmd`). However, if the contents of the notebooks have changed, it will use the restore key to load the latest version of `_freeze` that was run with the same Manifest.

The aim is to let each CI run access the most recently cached `_freeze` folder, to avoid re-rendering the same docs page on successive CI runs.

Closes #536

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -41,7 +41,10 @@ jobs:
         with:
           path: |
             ./_freeze/
-          key: ${{ runner.os }}-${{ hashFiles('**/Manifest.toml') }}
+          key: |
+            ${{ runner.os }}-${{ hashFiles('**/Manifest.toml') }}-${{ hashFiles('**/index.qmd') }}
+          restore-keys: |
+            ${{ runner.os }}-${{ hashFiles('**/Manifest.toml') }}
 
       - name: Render Quarto site
         run: quarto render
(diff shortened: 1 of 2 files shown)


mhauru (approved):
The code makes sense to me, but I don't understand the GA caching business well. Is there a way for us to see that it's working as intended?

penelopeysm:
Sadly, not until it's merged 🙃

Issue #547, opened 2024-11-03
Questions about AD support


gdalle:
I have a few questions about AD in Turing and [this docs page](https://turinglang.org/docs/tutorials/docs-10-using-turing-autodiff/).

> For ForwardDiff, pass `adtype=AutoForwardDiff(; chunksize)` to the sampler constructor. A `chunksize` of 0 permits the chunk size to be automatically determined. 

In ADTypes, the default `chunksize` is `nothing`, why adopt a different convention here? If you end up using DifferentiationInterface, `chunksize=0` will throw an error on every input.

> An additional argument can be provided to AutoReverseDiff to specify whether to to compile the tape only once and cache it for later use (false by default, which means no caching tape). [...] Compiled tapes should only be used if you are absolutely certain that the computation doesn’t change between different executions of your model. 

If I understand correctly, this holds for all tapes, compiled or not?

willtebbutt:
I'll let others address your point about ForwardDiff, however

```
If I understand correctly, this holds for all tapes, compiled or not?
```
Maybe there's some terminological confusion here. I think we need to distinguish between "caching" and "compiling". Control-flow related problems are associated to "caching" of a tape, while whether a tape is "compiled" or not (i.e. whether you've called `ReverseDiff.compile` on a tape that you're re-using each time you call `ReverseDiff.gradeint`) is irrelevant.

Is this also your mental model @gdalle ? If so, it might be that we should clarify the writing very slightly.

gdalle:
Yes that is the point I wanted to make.

gdalle:
Maybe this information (whether tape reuse is allowed) should go in the backend object from ADTypes.

willtebbutt:
Thanks for asking about this, and apologies for my confused response. Lets direct conversation to #548 now, rather than this issue.

PR #548, opened 2024-11-04
Update index.qmd

willtebbutt:
Addresses part of #547 .

@gdalle does this read more correctly?

### DIFF
--- a/tutorials/docs-10-using-turing-autodiff/index.qmd
+++ b/tutorials/docs-10-using-turing-autodiff/index.qmd
@@ -24,7 +24,8 @@ For `ReverseDiff`, pass `adtype=AutoReverseDiff()` to the sampler constructor. A
 
 
 
-Cached tapes should only be used if you are absolutely certain that the sequence of operations performed in your code does not change between different executions of your model.
+Pre-recorded tapes should only be used if you are absolutely certain that the sequence of operations performed in your code does not change between different executions of your model.
+
 Thus, e.g., in the model definition and all implicitly and explicitly called functions in the model, all loops should be of fixed size, and `if`-statements should consistently execute the same branches.
 For instance, `if`-statements with conditions that can be determined at compile time or conditions that depend only on fixed properties of the data will always execute the same branches during sampling (if the data is constant throughout sampling and, e.g., no mini-batching is used).
 However, `if`-statements that depend on the model parameters can take different branches during sampling; hence, the compiled tape might be incorrect.
--- a/tutorials/docs-10-using-turing-autodiff/index.qmd
+++ b/tutorials/docs-10-using-turing-autodiff/index.qmd
@@ -20,7 +20,8 @@ As of Turing version v0.30, the global configuration flag for the AD backend has
 Users can pass the `adtype` keyword argument to the sampler constructor to select the desired AD backend, with the default being `AutoForwardDiff(; chunksize=0)`.
 
 For `ForwardDiff`, pass `adtype=AutoForwardDiff(; chunksize)` to the sampler constructor. A `chunksize` of 0 permits the chunk size to be automatically determined. For more information regarding the selection of `chunksize`, please refer to [related section of `ForwardDiff`'s documentation](https://juliadiff.org/ForwardDiff.jl/dev/user/advanced/#Configuring-Chunk-Size).
-For `ReverseDiff`, pass `adtype=AutoReverseDiff()` to the sampler constructor. An additional argument can be provided to `AutoReverseDiff` to specify whether to to cache the tape only once and reuse it later use (`false` by default, which means no caching tape). This can substantially improve performance, but risks silently incorrect results if not used with care.
+For `ReverseDiff`, pass `adtype=AutoReverseDiff()` to the sampler constructor. An additional argument can be provided to `AutoReverseDiff` to specify whether to to cache the tape only once and reuse it later use (`false` by default, which means no caching). This can substantially improve performance, but risks silently incorrect results if not used with care.
+
 
 Cached tapes should only be used if you are absolutely certain that the sequence of operations performed in your code does not change between different executions of your model.
 Thus, e.g., in the model definition and all implicitly and explicitly called functions in the model, all loops should be of fixed size, and `if`-statements should consistently execute the same branches.
--- a/tutorials/docs-10-using-turing-autodiff/index.qmd
+++ b/tutorials/docs-10-using-turing-autodiff/index.qmd
@@ -19,7 +19,8 @@ While `Tracker` is still available, its use is discouraged due to a lack of acti
 As of Turing version v0.30, the global configuration flag for the AD backend has been removed in favour of [`AdTypes.jl`](https://github.com/SciML/ADTypes.jl), allowing users to specify the AD backend for individual samplers independently.
 Users can pass the `adtype` keyword argument to the sampler constructor to select the desired AD backend, with the default being `AutoForwardDiff(; chunksize=0)`.
 
-For `ForwardDiff`, pass `adtype=AutoForwardDiff(; chunksize)` to the sampler constructor. A `chunksize` of 0 permits the chunk size to be automatically determined. For more information regarding the selection of `chunksize`, please refer to [related section of `ForwardDiff`'s documentation](https://juliadiff.org/ForwardDiff.jl/dev/user/advanced/#Configuring-Chunk-Size).
+For `ForwardDiff`, pass `adtype=AutoForwardDiff(; chunksize)` to the sampler constructor. A `chunksize` of `nothing` permits the chunk size to be automatically determined. For more information regarding the selection of `chunksize`, please refer to [related section of `ForwardDiff`'s documentation](https://juliadiff.org/ForwardDiff.jl/dev/user/advanced/#Configuring-Chunk-Size).
+
 For `ReverseDiff`, pass `adtype=AutoReverseDiff()` to the sampler constructor. An additional keyword argument called `compile` can be provided to `AutoReverseDiff`. It specifies whether to pre-record the tape only once and reuse it later (`compile` is set to `false` by default, which means no pre-recording). This can substantially improve performance, but risks silently incorrect results if not used with care.
 
 
--- a/tutorials/docs-10-using-turing-autodiff/index.qmd
+++ b/tutorials/docs-10-using-turing-autodiff/index.qmd
@@ -20,7 +20,8 @@ As of Turing version v0.30, the global configuration flag for the AD backend has
 Users can pass the `adtype` keyword argument to the sampler constructor to select the desired AD backend, with the default being `AutoForwardDiff(; chunksize=0)`.
 
 For `ForwardDiff`, pass `adtype=AutoForwardDiff(; chunksize)` to the sampler constructor. A `chunksize` of 0 permits the chunk size to be automatically determined. For more information regarding the selection of `chunksize`, please refer to [related section of `ForwardDiff`'s documentation](https://juliadiff.org/ForwardDiff.jl/dev/user/advanced/#Configuring-Chunk-Size).
-For `ReverseDiff`, pass `adtype=AutoReverseDiff()` to the sampler constructor. An additional argument can be provided to `AutoReverseDiff` to specify whether to to cache the tape only once and reuse it later use (`false` by default, which means no caching). This can substantially improve performance, but risks silently incorrect results if not used with care.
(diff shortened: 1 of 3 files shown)


tutorials/docs-10-using-turing-autodiff/index.qmd
```diff
@@ -20,11 +20,11 @@ As of Turing version v0.30, the global configuration flag for the AD backend has
...
 
 For `ForwardDiff`, pass `adtype=AutoForwardDiff(; chunksize)` to the sampler constructor. A `chunksize` of 0 permits the chunk size to be automatically determined. For more information regarding the selection of `chunksize`, please refer to [related section of `ForwardDiff`'s documentation](https://juliadiff.org/ForwardDiff.jl/dev/user/advanced/#Configuring-Chunk-Size).
-For `ReverseDiff`, pass `adtype=AutoReverseDiff()` to the sampler constructor. An additional argument can be provided to `AutoReverseDiff` to specify whether to to compile the tape only once and cache it for later use (`false` by default, which means no caching tape). Be aware that the use of caching in certain types of models can lead to incorrect results and/or errors.
+For `ReverseDiff`, pass `adtype=AutoReverseDiff()` to the sampler constructor. An additional argument can be provided to `AutoReverseDiff` to specify whether to to cache the tape only once and reuse it later use (`false` by default, which means no caching tape). This can substantially improve performance, but risks silently incorrect results if not used with care.
```
willtebbutt, line 23:
```suggestion
For `ReverseDiff`, pass `adtype=AutoReverseDiff()` to the sampler constructor. An additional argument can be provided to `AutoReverseDiff` to specify whether to to cache the tape only once and reuse it later use (`false` by default, which means no caching). This can substantially improve performance, but risks silently incorrect results if not used with care.

```

gdalle:
I think there might still be a misunderstanding here. The keyword argument `AutoReverseDiff(; compile)` only controls whether the tape is optimized after being recorded. It doesn't control whether a recorded tape is only used once, or reused across function calls (inducing possible control flow problems).
Tape reuse is instead determined by how the `ADGradient` is created, and whether or not an example vector `x` is passed to it. See these key lines in LogDensityProblemsAD.jl:

https://github.com/tpapp/LogDensityProblemsAD.jl/blob/e3401f21b5a065df0d5de38b37fad0e6650618f3/ext/LogDensityProblemsADReverseDiffExt.jl#L45-L56

So the real question is:

- in which places does Turing "prepare" the `ADGradient` with an example `x`
- can the user of the library do anything about it, if they know that their loglikelihood has value-dependent control flow?

torfjelde:
> in which places does Turing "prepare" the ADGradient with an example x

Typically this occurs in the initial step of the sampler, i.e. once (there are exceptions, but in those cases you can't really do much better in our case).

> can the user of the library do anything about it, if they know that their loglikelihood has value-dependent control flow?

Unfortunately not for ReverseDiff.jl in compiled mode (they can however just not use compiled mode).

tutorials/docs-10-using-turing-autodiff/index.qmd
```diff
@@ -20,11 +20,12 @@ As of Turing version v0.30, the global configuration flag for the AD backend has
...
-Thus, e.g., in the model definition and all im- and explicitly called functions in the model all loops should be of fixed size, and `if`-statements should consistently execute the same branches.
-For instance, `if`-statements with conditions that can be determined at compile time or conditions that depend only on the data will always execute the same branches during sampling (if the data is constant throughout sampling and, e.g., no mini-batching is used).
+
+Cached tapes should only be used if you are absolutely certain that the sequence of operations performed in your code does not change between different executions of your model.
+Thus, e.g., in the model definition and all implicitly and explicitly called functions in the model, all loops should be of fixed size, and `if`-statements should consistently execute the same branches.
+For instance, `if`-statements with conditions that can be determined at compile time or conditions that depend only on fixed properties of the data will always execute the same branches during sampling (if the data is constant throughout sampling and, e.g., no mini-batching is used).
```
torfjelde, line 28:
```suggestion
For instance, `if`-statements with conditions that can be determined at compile time or conditions that depend only on fixed properties of the model, e.g. fixed data.
```

I don't think there's much point mentioning minibatching here, as there's no "easy to use" support for this in Turing.jl and so it's really not something people do much of (don't think I've ever seen anyone actually do this in applications with Turing.jl).

tutorials/docs-10-using-turing-autodiff/index.qmd
```diff
@@ -20,11 +20,12 @@ As of Turing version v0.30, the global configuration flag for the AD backend has
 Users can pass the `adtype` keyword argument to the sampler constructor to select the desired AD backend, with the default being `AutoForwardDiff(; chunksize=0)`.
 
 For `ForwardDiff`, pass `adtype=AutoForwardDiff(; chunksize)` to the sampler constructor. A `chunksize` of 0 permits the chunk size to be automatically determined. For more information regarding the selection of `chunksize`, please refer to [related section of `ForwardDiff`'s documentation](https://juliadiff.org/ForwardDiff.jl/dev/user/advanced/#Configuring-Chunk-Size).
```
torfjelde, line 22:
```suggestion
For `ForwardDiff`, pass `adtype=AutoForwardDiff(; chunksize)` to the sampler constructor. A `chunksize` of `nothing` permits the chunk size to be automatically determined. For more information regarding the selection of `chunksize`, please refer to [related section of `ForwardDiff`'s documentation](https://juliadiff.org/ForwardDiff.jl/dev/user/advanced/#Configuring-Chunk-Size).

```

torfjelde: Maybe also add this, as @gdalle pointed out, this is not actually the case anymore (this was a left-over thing from using LogDensityProblemsAD.jl I believe, but this changed when it all moved to ADTypes.jl backed)

torfjelde (commented):
Added some comments 👍

gdalle:
> Unfortunately not for ReverseDiff.jl in compiled mode (they can however just not use compiled mode).

@torfjelde I think that's the key confusion between us. My understanding is that, _regardless of whether it is compiled or not_, the tape only captures execution for one version of the control flow. Thus, the tape always becomes invalid if the control flow changes (e.g. different branch of an `if` statement). Isn't that how it goes?

Source: <https://juliadiff.org/ReverseDiff.jl/dev/api/#The-AbstractTape-API>

devmotion:
The tape is purely internal and neither stored nor reused if `compile = Val(false)`.

> Tape reuse is instead determined by how the ADGradient is created, and whether or not an example vector x is passed to it

This is wrong. The tape is only reused if `compile = Val(true)`: https://github.com/tpapp/LogDensityProblemsAD.jl/blob/e3401f21b5a065df0d5de38b37fad0e6650618f3/ext/LogDensityProblemsADReverseDiffExt.jl#L51-L56

gdalle:
Okay then IIUC it's a matter of inconsistency between
- the "compile" kwarg here and in LogDensityProblemsAD (which determines whether a tape is reused)
- the compilation as defined by ReverseDiff, which is about speeding up an existing tape (and disconnected from the reuse issue)

gdalle:
Note that in ADTypes, the "compile" argument to AutoReverseDiff is defined ambiguously too. So we should perhaps add more details to the struct, something like AutoReverseDiff(tape=true, compile=false)?

gdalle:
I'd love for you to chime in here @devmotion @torfjelde: https://github.com/SciML/ADTypes.jl/issues/91

gdalle:
> I'm not too opinionated about this, as compilation without caching seems somewhat useless? Are there scenarios where you'd like to do that?

Indeed you can't compile a tape you never record in the first place. In any case, I think the ambiguous terminology was fixed by https://github.com/SciML/ADTypes.jl/issues/91. It's just a shame that the word "compile" was chosen instead of "record", given how both are used in ReverseDiff's documentation. But it's a sunk cost now.

tutorials/docs-10-using-turing-autodiff/index.qmd
```diff
@@ -20,11 +20,12 @@ As of Turing version v0.30, the global configuration flag for the AD backend has
...
 
 For `ForwardDiff`, pass `adtype=AutoForwardDiff(; chunksize)` to the sampler constructor. A `chunksize` of 0 permits the chunk size to be automatically determined. For more information regarding the selection of `chunksize`, please refer to [related section of `ForwardDiff`'s documentation](https://juliadiff.org/ForwardDiff.jl/dev/user/advanced/#Configuring-Chunk-Size).
-For `ReverseDiff`, pass `adtype=AutoReverseDiff()` to the sampler constructor. An additional argument can be provided to `AutoReverseDiff` to specify whether to to compile the tape only once and cache it for later use (`false` by default, which means no caching tape). Be aware that the use of caching in certain types of models can lead to incorrect results and/or errors.
+For `ReverseDiff`, pass `adtype=AutoReverseDiff()` to the sampler constructor. An additional argument can be provided to `AutoReverseDiff` to specify whether to to cache the tape only once and reuse it later use (`false` by default, which means no caching). This can substantially improve performance, but risks silently incorrect results if not used with care.
```
gdalle, line 23:
```suggestion
For `ReverseDiff`, pass `adtype=AutoReverseDiff()` to the sampler constructor. An additional keyword argument called `compile` can be provided to `AutoReverseDiff`. It specifies whether to pre-record the tape only once and reuse it later (`compile` is set to `false` by default, which means no pre-recording). This can substantially improve performance, but risks silently incorrect results if not used with care.

```

tutorials/docs-10-using-turing-autodiff/index.qmd
```diff
@@ -20,11 +20,12 @@ As of Turing version v0.30, the global configuration flag for the AD backend has
...
 
-Compiled tapes should only be used if you are absolutely certain that the computation doesn't change between different executions of your model.
-Thus, e.g., in the model definition and all im- and explicitly called functions in the model all loops should be of fixed size, and `if`-statements should consistently execute the same branches.
-For instance, `if`-statements with conditions that can be determined at compile time or conditions that depend only on the data will always execute the same branches during sampling (if the data is constant throughout sampling and, e.g., no mini-batching is used).
+
+Cached tapes should only be used if you are absolutely certain that the sequence of operations performed in your code does not change between different executions of your model.
```
gdalle, line 26:
```suggestion
Pre-recorded tapes should only be used if you are absolutely certain that the sequence of operations performed in your code does not change between different executions of your model.

```

willtebbutt:
Thank you all for raising this problem / helping to fix it -- it's good that we've gotten to the bottom of this! I'm basically happy with this. @gdalle and @torfjelde could take a final look at this, and let me know whether you are happy.

torfjelde:
> It's just a shame that the word "compile" was chosen instead of "record", given how both are used in ReverseDiff's documentation. But it's a sunk cost now.

I agree with that 😕

gdalle (approved):

torfjelde (approved):
LGTM! Added one final comment that would be nice to shoehorn in here since we're already making alterations, but looks good:)

tutorials/docs-10-using-turing-autodiff/index.qmd
```diff
@@ -62,4 +66,4 @@ c = sample(
 
 Generally, reverse-mode AD, for instance `ReverseDiff`, is faster when sampling from variables of high dimensionality (greater than 20), while forward-mode AD, for instance `ForwardDiff`, is more efficient for lower-dimension variables. This functionality allows those who are performance sensitive to fine tune their automatic differentiation for their specific models.
 
-If the differentiation method is not specified in this way, Turing will default to using whatever the global AD backend is. Currently, this defaults to `ForwardDiff`.
\ No newline at end of file
+If the differentiation method is not specified in this way, Turing will default to using whatever the global AD backend is. Currently, this defaults to `ForwardDiff`.
```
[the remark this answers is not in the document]
gdalle: When you fully switch to DI you can also use `DifferentiationInterfaceTest.benchmark_differentiation` for this ;)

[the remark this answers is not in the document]
willtebbutt: This is a good point. @torfjelde would it make more sense to just link to TuringBenchmarking.jl, rather than copying + pasting in a code snippet which may become outdated? Alternatively, is there a simple way to put example benchmarking code in here which runs whenever the docs are built?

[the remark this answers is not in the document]
willtebbutt: Thanks for reminding me about this @torfjelde . I've made the changes requested, so I'm going to mark this conversation closed. I'll get CI passing and then merge.

willtebbutt:
@penelopeysm it looks like my PR has the potential to cause merge conflicts with #559 . Would you like me to wait until you've merged yours before merging this, or are you happy to resolve any conflicts that arise?

penelopeysm:
I'm happy to sort it out :)

PR #549, opened 2024-11-06
Replace truncated(d, 0, Inf) -> truncated(d; lower=0)

penelopeysm:
cf. https://github.com/JuliaStats/Distributions.jl/issues/1910


devmotion (approved):
Ah, good catch! Quit unfortunate that I had missed these when updating the docs to the kwargs version.

PR #552, opened 2024-11-15
Add `import Pkg` for ease of copy-pasting

penelopeysm:
The docs on contributing say to run `Pkg.test`, but it's annoying to copy-paste this and get hit with Pkg not defined.

This fixes it so that the test command can be copied as a one-liner into a fresh Julia REPL.

Also specifies that you need to use `julia --project=.` from the top level, as I don't think there's any harm in being explicit.


mhauru (approved):
LGTM

PR #559, opened 2024-11-25
Update to Turing v0.35

### LINKED ISSUE #521: Replace `Tracker` with `Mooncake` in Bayesian neural networks example
Let's depreciate the support for `Tracker` (we mention `Tracker`in various autodiff doc pages) since it is no longer actively maintained and is known to be problematic for some Bijectors.

### LINKED ISSUE #541: Prepare update for Turing v0.35
## TODO (in this order)

 - [x] Release Turing ~~0.35.0~~ we're at 0.35.2 now
 - [x] Bump min Julia version on intro page
 - ~~Verify that changelog is updated~~ (Can't do this until it's merged)
 - [x] https://github.com/TuringLang/docs/issues/521
 - [x] Get rid of any other mention of Tracker, add Mooncake where necessary
 - [x] Port #539

penelopeysm:
~~I suspect this might yet crash and burn because of the Bijectors stuff, but have to try it.~~

It seems to be fine although we are restricted to Julia 1.10 and Turing 0.35.1 (not 0.35.2), as the latter bumps Bijectors compat to 0.14 which causes a bunch of dependency constraints that I didn't dig into. My suspicion is that all of the upgrades (Julia -> 1.11 and Turing -> 0.35.2) are contingent on https://github.com/TuringLang/Bijectors.jl/pull/341 being merged.

This PR also updates the Bayesian neural network tutorial to work with Mooncake. It didn't work out of the box, I think that this was because of mixed-precision floating point arithmetic. Wrapping the neural network parameters in `Lux.f32` (see https://lux.csail.mit.edu/stable/api/Lux/utilities#Updating-Floating-Point-Precision) allowed AD and inference to run locally.

I'm fairly confident that the docs should build now.

Closes #521.

Closes #541.


tutorials/03-bayesian-neural-network/index.qmd
```diff
@@ -189,7 +189,7 @@ const nn = StatefulLuxLayer{true}(nn_initial, nothing, st)
     parameters ~ MvNormal(zeros(nparameters), Diagonal(abs2.(sigma .* ones(nparameters))))
 
     # Forward NN to make predictions
-    preds = Lux.apply(nn, xs, vector_to_parameters(parameters, ps))
+    preds = Lux.apply(nn, xs, f32(vector_to_parameters(parameters, ps)))
```
penelopeysm, line 192: @willtebbutt I had to change this line in the Bayesian neural net tutorial to get it to work with Mooncake. Without this, I think it was doing some f32 * f64 matrix multiplication for which `build_rrule` failed: https://github.com/TuringLang/docs/actions/runs/12017000042/job/33498395092#step:8:537 Happy to create an issue at a later point, just wanted to run this past you to make sure it's a sensible fix.

willtebbutt: Thanks for highlighting this @penelopeysm , and very good to know that this is working well. I think I'm going to have to generalise my implementation of the rule for `LuxLib.Impl.matmul` to handle mixed precision inputs, which is not something I thought would be necessary!

PR #527, opened 2024-10-01
Add concurrency key to publish/preview workflows

penelopeysm:
This PR adds the concurrency key `docs` to both publish and preview workflows, preventing them from running concurrently. This will help us avoid weird race conditions like the one described in #526, which came about because the preview workflow overwrote changes from the publish workflow.

Additionally, the publish workflow is also given `cancel-in-progress: true` meaning that if a preview workflow is running, starting a publish workflow will cancel the preview workflow.

This is probably OK because preview workflows run on PRs, and if somebody pushes to main (which triggers the publish workflow), the PR has to incorporate the new changes anyway.

The only downside to this is that there can only be one queued job at any time. So if (for example) three people are working on PRs at the same time and trigger preview workflows, one workflow would run, one workflow would be queued, and one of them would be cancelled. That scenario seems quite unlikely given the level of contributions to the docs, though, and can be fixed fairly easily by manually rerunning the workflow. (This is a weird limitation of GHA, and apparently a fix for this is on GitHub's roadmap, see https://github.com/orgs/community/discussions/41518.)

Closes #526.

CI is failing because I manually cancelled the preview workflow on this PR (it doesn't serve any purpose).

PR #563, opened 2024-11-26
Fix GHA stuff

penelopeysm:
This PR makes a few changes to GitHub Actions workflows:

1. Separates the preview workflow from the remove-preview workflow.
   
   Whenever a PR was merged to master, the remove-preview workflow would fire. However, the publish workflow on master would also fire. The changes in #527 meant that the publish workflow would cancel the remove-preview workflow. The result is that we have a buildup of PR previews.
   
   This fixes it by making the remove-preview workflow a separate one entirely, so that its running is unaffected by other workflows.

2. Modifies the version check workflow to not use `continue-on-error: true`. This causes the workflow to always report a green tick, even if the step did have an error. Instead I use `always()` on the subsequent steps.

3. Slightly improves the version checking script.

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -5,7 +5,6 @@ on:
     types:
       - opened
       - synchronize
-      - closed
 
 concurrency:
   group: docs
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -10,6 +10,10 @@ on:
 concurrency:
   group: docs
 
+permissions:
+  contents: write
+  pull-requests: write
+
 jobs:
   build-and-preview:
     if: github.event.action == 'opened' || github.event.action == 'synchronize'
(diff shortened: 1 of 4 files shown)


penelopeysm:
The effects of (2) are already visible 😄 it is failing because it can't resolve an env with Turing v0.35.2, as expected, see https://github.com/TuringLang/docs/pull/559#issue-2691902578.

shravanngoswamii (approved):
Looks good, thank you, @penelopeysm!

Fixing these issues in GHA is really great. I was aware of the PR-preview removal issue, but we decided to ignore it when merging this repo into Quarto. At that time, we were mainly focused on addressing internal documentation problems, so the GHA workflows were written to just work, but were not fully optimized or cleaned up. It's really helpful to have this addressed now—thanks for your efforts!

penelopeysm:
Thanks @shravanngoswamii! To be honest, I'm pretty sure that this is still not quite right. There's probably some weird behaviour that I haven't figured out. But with GHA I always think it's fine to merge something imperfect since it can't really cause any bugs in the actual code 😅

shravanngoswamii:
> To be honest, I'm pretty sure that this is still not quite right. There's probably some weird behaviour that I haven't figured out. But with GHA I always think it's fine to merge something imperfect since it can't really cause any bugs in the actual code 😅

Let's see, we can always change it if it creates any issue, experiments takes the product near perfection😅!

PR #574, opened 2025-01-16
Normalise line endings to LF

penelopeysm:
This PR:

- runs `dos2unix` on every text file in the repository to convert any stray CRLFs to LFs,
- adds a `.gitattributes` file to keep them the same under future edits.


.gitattributes
```diff
@@ -0,0 +1,9 @@
+################
+# Line endings #
+################
+*     text=auto
+
+###################
+# GitHub Linguist #
+###################
+*.qmd linguist-language=Markdown
```
mhauru, line 9: It seemed to me like they were already getting Markdown syntax highlighting. Was there something in particular that prompted this?

mhauru (approved):
Beautiful

mhauru:
Do you know what's up with the failing version consistency check, "Unsatisfiable requirements detected"? Seemed to pass a couple of weeks ago.

penelopeysm:
(I suppose that this is the problem in a system that uses conservative upper bounds + frequent minor releases)

PR #579, opened 2025-01-16
Turing v0.36 and removal of Pathfinder code

mhauru:
This PR makes two unrelated changes:
1. It upgrades Turing to v0.36, and correspondingly updates instructions for how to use the Gibbs sampler.
1. It removes the code example for how to use Pathfinder, and thus the docs environment's dependency on Pathfinder.

These are in the same PR, because doing the first one made us realise that having Pathfinder as dependency is annoying: We couldn't upgrade Turing.jl to v0.36 because there was not yet a version of Pathfinder that included it in its compat, and we thus couldn't build docs for the latest version of our own package.


penelopeysm:
Oh no, something about the new Gibbs is failing 😓

penelopeysm (approved):
If you're happy I'm happy

mhauru:
Thanks for fixing this!

PR #582, opened 2025-01-24
Variable transformations

penelopeysm:
This PR adds a series of new docs on:

- why variables might need to be transformed;
- how this is accounted for mathematically with the Jacobian term;
- basic API and functionality of Bijectors.jl;
- higher-level usage of transforms in DynamicPPL and Turing.

Comments are welcome:)

PS. I know that there is some subtlety with the Turing.jl HMC sampling section because the print statements don't actually happen during the sampling, they occur when `values_as_in_model` is called (post-sampling) to collect the values. I figured it's ok to gloss over that, though, because it's still true that HMC sampling occurs entirely in unconstrained space.


developers/transforms/distributions/index.qmd
```diff
@@ -0,0 +1,314 @@
...
+```
+
+The probability density function for the normal distribution with mean 0 and standard deviation 1 is
+
+$$p(x) = \frac{1}{\sqrt{2\pi}} \exp{\left(-\frac{x^2}{2}\right)},$$
+
+so we could also have calculated this manually using:
+
+```{julia}
+log(1 / sqrt(2π) * exp(-samples[1]^2 / 2))
```
willtebbutt, line 53: I know you've just got this here as an example, but would you be up for adding a footnote that warns people against computing the logpdf in this way? i.e. they really ought to simplify the expression first to avoid ever exponentiating -x^2/2

developers/transforms/distributions/index.qmd
```diff
@@ -0,0 +1,314 @@
...
+For example, in the normal distribution, half of the probability density is between $-\infty$ and $0$, and half is between $0$ and $\infty$.
+When exponentiated (i.e. in the log-normal distribution), the first half of the density is mapped to the interval $(0, 1)$, and the second half to $(1, \infty)$.
+
+This 'explanation' on its own does not really mean much, though.
+A perhaps more useful approach is to not talk about _probability densities_, but instead to make it more concrete by relating them to actual _probabilities_.
+If we think about the normal distribution as a continuous curve, what the probability density function $p(x)$ really tells us is that: for any two points $a$ and $b$ (where $a \leq b$), the probability of drawing a sample between $a$ and $b$ is the corresponding area under the curve, i.e.
+
+$$\int_a^b p(x) \, \mathrm{d}x.$$
+
+For example, if $(a, b) = (-\infty, \infty)$, then the probability of drawing a sample from the entire distribution is 1.
```
willtebbutt, line 118:
```suggestion
For example, if $(a, b) = (-\infty, \infty)$, then the probability of drawing a sample between $a$ and $b$ is 1.
```
grammar suggestion. Feel free to ignore if you don't like.

developers/transforms/distributions/index.qmd
```diff
@@ -0,0 +1,314 @@
...
+
+and check that it works:
+
+```{julia}
+println("Sample   : $(samples_lognormal[1])")
+println("Expected : $(logpdf(LogNormal(), samples_lognormal[1]))")
+println("Actual   : $(logpdf(MyLogNormal(), samples_lognormal[1]))")
+```
+
+The same process can be applied to _any_ kind of transformation.
```
willtebbutt, line 166: Would you be open to restricting this a little? i.e. this doesn't work for non-invertible transformations.

developers/transforms/distributions/index.qmd
```diff
@@ -0,0 +1,314 @@
...
+
+(where we used the product rule), and
+
+$$\frac{\partial x_2}{\partial y_1} = \frac{1}{2\pi} \left(\frac{1}{1 + (y_2/y_1)^2}\right) \left(-\frac{y_2}{y_1^2}\right),$$
+
+(where we used the chain rule, and the derivative $\mathrm{d}(\arctan(a))/\mathrm{d}a = 1/(1 + a^2)$).
+
+Putting together the Jacobian matrix, we have:
+
+$$\mathcal{J} = \begin{pmatrix}
```
willtebbutt, line 280:
```suggestion
$$\mathbfl{J} = \begin{pmatrix}
```
for consistency with the previous section

developers/transforms/distributions/index.qmd
```diff
@@ -0,0 +1,314 @@
...
+-y_1 x_1 & -y_2 x_1 \\
+-cy_2/y_1^2 & c/y_1 \\
+\end{pmatrix},$$
+
+where $c = [2\pi(1 + (y_2/y_1)^2)]^{-1}$.
+The determinant of this matrix is
+
+$$\begin{align}
+\det(\mathcal{J}) &= -cx_1 - cx_1(y_2/y_1)^2 \\
```
willtebbutt, line 289:
```suggestion
\det(\mathbf{J}) &= -cx_1 - cx_1(y_2/y_1)^2 \\
```
same again

developers/transforms/distributions/index.qmd
```diff
@@ -0,0 +1,314 @@
...
+&= -cx_1\left[1 + \left(\frac{y_2}{y_1}\right)^2\right] \\
+&= -\frac{1}{2\pi} x_1 \\
+&= -\frac{1}{2\pi}\exp{\left(-\frac{y_1^2}{2}\right)}\exp{\left(-\frac{y_2^2}{2}\right)},
+\end{align}$$
+
+Coming right back to our probability density, we have that
+
+$$\begin{align}
+q(y_1, y_2) &= p(x_1, x_2) \cdot |\det(\mathcal{J})| \\
```
willtebbutt, line 298:
```suggestion
q(y_1, y_2) &= p(x_1, x_2) \cdot |\det(\mathbf{J})| \\
```
same as before

willtebbutt (commented):
Just minor stuff. I like these docs a lot!

developers/transforms/distributions/index.qmd
willtebbutt: I like this!

developers/transforms/bijectors/index.qmd
```diff
@@ -0,0 +1,311 @@
...
+histogram(samples_untransformed, bins=0:0.1:5; xlims=(0, 5))
+```
+
+We can check the mean of the samples too, to see that it is what we expect:
+
+```{julia}
+println("expected mean: $(exp(0 + (1^2/2)))")
+println("  actual mean: $(mean(samples_untransformed))")
+```
+
```
mhauru, line 250: Would it be helpful to print out the number of rejections when using the bijector, just to rub in the point that it's zero? Either here in or in the next section. It's not a conclusive proof that using a bijector is better, but it's a strong hint.

developers/transforms/dynamicppl/index.qmd
```diff
@@ -0,0 +1,399 @@
...
+```{julia}
+new_dist_x = truncated(Normal(); lower=vi_linked[vn_m])
+new_f_inv = DynamicPPL.invlink_transform(new_dist_x)
+```
+
+and if we apply this to the internal representation of `x`:
+
+```{julia}
+new_f_inv(DynamicPPL.getindex_internal(vi_linked, vn_x))
+```
```
mhauru, line 399: I feel like this could use a final sentence pointing out that the value is the same we saw before.

Also, I feel rude leaving behind such a friendly text that you've bonded with over three articles without saying goodbye.

mhauru (approved):
This is great! I had a few small comments, but they are minor enough that I'm happy to merge as-is if need be.

developers/transforms/dynamicppl/index.qmd
```diff
@@ -0,0 +1,399 @@
...
+However, if it was linked, then the Jacobian term would be appropriately included, making sure that sampling proceeds correctly.
+
+## Why do we need to do this at runtime?
+
+Given that we know whether a `VarInfo` is linked or not, one might wonder why we need both `from_internal_transform` and `from_linked_internal_transform` at the point where the model is evaluated.
+Could we not, for example, store the required transformation inside the `VarInfo` when we link it, and simply reuse that each time?
+
+That is, why can't we just do
+
+```{mermaid}
```
mhauru, line 278: Forgot the most important comment: The mermaid diagrams render funny on the preview page: https://turinglang.org/docs/pr-previews/582/developers/transforms/dynamicppl/

developers/transforms/bijectors/index.qmd
penelopeysm: I'm a bit wary of that metric, because you could be sampling very inefficiently in unconstrained space, and thus end up getting bad results despite not rejecting any samples. It's a bit like the MH proposal width thing, you could make it be very small and thus not get any rejections, but it doesn't necessarily mean that it's better.

(I did have to rework this section a couple of times until I found a satisfactory way to measure which method was better 🥲)

penelopeysm: But I have now edited it to print out the above + include the explanation in my comment 😄

developers/transforms/dynamicppl/index.qmd
penelopeysm: Oh no, I remember spending ages trying to figure out weird things with Quarto + Mermaid.

mhauru: It's dangerous to go alone! Take this. 💻

penelopeysm: There was some offending CSS code:

https://github.com/TuringLang/docs/blob/523428a5a8179bcced2254bd5f708b71288a0b11/theming/styles.css#L35-L38

Who says there isn't web dev on Turing projects!

penelopeysm:
I'll leave this up a few more days to see if anybody has more input before merging:)

PR #589, opened 2025-03-23
Turing 0.37

penelopeysm:
- Updates version of Turing.jl to 0.37
- Makes accompanying changes to the code to adjust for this, mostly around `.~` and also fixing imports of Bijectors


sunxd3 (approved):
The changed parts are all good, and the preview doesn't error. So it looks good to me!

PR #592, opened 2025-03-24
Fix differential equation tutorial not converging

### LINKED ISSUE #588: ODE example does not converge to true solution.
https://turinglang.org/docs/tutorials/bayesian-differential-equations/index.html

One of the chains seems to be stuck in a local minimum. This should be fixed or at least mentioned in the docs.

penelopeysm:
Closes #588 by reducing the amount of random noise in the synthetic data, and also making the priors in the model more strongly peaked on the ground truth. (I suppose that this is a bit of a cheat and ideally we'd sample more, but the docs already take a long time to run.)

Before: https://turinglang.org/docs/tutorials/bayesian-differential-equations/
After: https://turinglang.org/docs/pr-previews/592/tutorials/bayesian-differential-equations/

### DIFF
--- a/tutorials/bayesian-differential-equations/index.qmd
+++ b/tutorials/bayesian-differential-equations/index.qmd
@@ -79,7 +79,7 @@ To make the example more realistic we add random normally distributed noise to t
 
 ```{julia}
 sol = solve(prob, Tsit5(); saveat=0.1)
-odedata = Array(sol) + 0.8 * randn(size(Array(sol)))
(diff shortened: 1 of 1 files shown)


sunxd3:
Categorically, I have no objection to the updates here.

Maybe you could add a sentence on prior selection, it can be something like "we are choosing prior close to the ground truth". I think it might make it more clear what we are trying to demonstrate here. This is not related to the changes made in this PR, just something crosses my mind.

PR #595, opened 2025-03-28
Formalise AD integration status, rewrite AD page correspondingly

penelopeysm:
This is an initial attempt to put down in words what @willtebbutt and I think our approach to integrating AD backends should be going forward.


usage/automatic-differentiation/index.qmd
```diff
@@ -12,34 +12,129 @@ using Pkg;
...
+
+**Tier 2** indicates some level of confidence on our side that the AD library will work, because it is included as part of DynamicPPL's continuous integration (CI) tests.
+If you find that a Tier 2 backend does not work with Turing, you are welcome to submit an issue, and we will try to look into it.
+Note, however, that this does not imply that we take responsibility for ensuring that any given model will work with these backends.
+This may be either due to upstream bugs / limitations (which exist even for ForwardDiff), or simply because of time constraints.
+However, if there are workarounds that can be implemented in Turing to make the backend work, we will try to do so.
+
+**Tier 3** is the same as Tier 2, but in addition to that, we formally also take responsibility for ensuring that the backend works with Turing models.
+If you submit an issue about using Turing with a Tier 3 library, we will actively try to make it work.
+Realistically, this is only possible for AD backends that are actively maintained by somebody on the Turing team, such as Mooncake.
```
willtebbutt, line 78: What are the limits to how far we're willing to take this? Per our discussion yesterday, if someone does something really non-differentiable (e.g. a custom `ccall`), we're not going to try and add support for their proposal.

Maybe "we will actively try to make it work" could be extended to say how we'll try to make it work? e.g. if someone encounters a bug, we'll fix it, but if they're doing something unusual we might suggest a more standard way to go about it that avoids the problem they're seeing entirely.

usage/automatic-differentiation/index.qmd
```diff
@@ -65,14 +160,69 @@ c = sample(
...
+
+```{julia}
+@model function forwarddiff_working2(::Type{T}=Float64) where T
+    x = T[0.0, 1.0]
+    a ~ Normal()
+    x[1] = a
+    b ~ MvNormal(x, I)
+end
+sample(forwarddiff_working2(), NUTS(; adtype=AutoForwardDiff()), 10)
+```
```
willtebbutt, line 214: Would it be helpful for users to make it clear that the second option here is highly preferable to the first in general, and that the first should only be used if the second doesn't work for some reason?

willtebbutt (commented):
Some thoughts.

usage/automatic-differentiation/index.qmd
penelopeysm: Yes, definitely, and a link to https://discourse.julialang.org/t/vector-real-vector-float64-methoderror/25926/5 might also be helpful (it helped me back in the day)

usage/automatic-differentiation/index.qmd
```diff
@@ -12,34 +12,129 @@ using Pkg;
...
+
+There are two aspects to choosing an AD backend: firstly, what backends are available; and secondly, which backend is best for your model.
+
+### Usable AD Backends
+
+Turing.jl uses the functionality in [DifferentiationInterface.jl](https://github.com/JuliaDiff/DifferentiationInterface.jl) ('DI') to interface with AD libraries in a unified way.
+Thus, in principle, any AD library that has integrations with DI can be used with Turing; you should consult the [DI documentation](https://juliadiff.org/DifferentiationInterface.jl/DifferentiationInterface/stable/) for an up-to-date list of compatible AD libraries.
+
+Note, however, that not all AD libraries in there are tested on Turing models.
+Thus, it is likely that some of them will either error (because they don't know how to differentiate through Turing's code), or might silently give incorrect results.
```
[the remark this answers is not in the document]
penelopeysm: I can't see why I would be against it 😄

usage/automatic-differentiation/index.qmd
mhauru: Maybe we could add a paragraph above or below the list of tiers explaining what it means that a backend works with Turing? It's mostly relevant for Tier 3, but it could be a good clarification more generally too. Something like

> When we say that an AD backend works with Turing, we mean that it is able to differentiate any Turing model that depends only on Turing and some common Julia standard library modules such as LinearAlgebra. Note that a Turing model can include arbitrary Julia code, which can involve code dependencies on other packages such as differential equation solvers or external calls using `ccall` (should something else be added to the list of exclusions?). If a Tier 2 or 3 AD backend fails on a Turing model because of such features, we may still be able to help you out in some cases, but we may also consider the problem to be outside our control or scope.

Good to also keep in mind that while it's nice to be clear and explicit about our thinking, it's not a legal contract and we don't have to be suuuuper precise about our wording on what we commit to fixing. It's all still subject to the usual uncertainties of academic funding and time anyway.

penelopeysm:
I feel like most of this is obsolete now given ADTests. I'll make a new PR for this.

PR #599, opened 2025-05-09
Update to Turing 0.38

penelopeysm:
`generated_quantities` was deprecated quite a few DPPL versions ago, so I replaced it with `returned`, and also added a section talking about `:=`.

Also replaced TuringBenchmarking with in-house DPPL benchmarking.

Also removed Zygote from the list of 'officially supported'. Technically, this would have been done in #595, but I'm slow on getting that merged.

Also added a reference to ADTests.

Also fixed some other dead links.


usage/automatic-differentiation/index.qmd
```diff
@@ -33,8 +33,6 @@ For instance, `if`-statements with conditions that can be determined at compile
 However, `if`-statements that depend on the model parameters can take different branches during sampling; hence, the compiled tape might be incorrect.
 Thus you must not use compiled tapes when your model makes decisions based on the model parameters, and you should be careful if you compute functions of parameters that those functions do not have branching which might cause them to execute different code for different values of the parameter.
 
-For `Zygote`, pass `adtype=AutoZygote()` to the sampler constructor.
-
```
mhauru, line 37: We could add a similar note for Mooncake. Optional.

usage/tracking-extra-quantities/index.qmd
```diff
@@ -0,0 +1,152 @@
...
+```{julia}
+#| echo: false
+#| output: false
+using Pkg;
+Pkg.instantiate();
+```
+
+Often, there are quantities in models that we might be interested in viewing the values of, but which are not random variables in the model that are explicitly drawn from a distribution.
+
+As a motivating example, the most natural parameterization for a model might not the most computationally feasible.
```
mhauru, line 18:
```suggestion
As a motivating example, the most natural parameterization for a model might not be the most computationally feasible.
```

usage/tracking-extra-quantities/index.qmd
```diff
@@ -0,0 +1,152 @@
...
+```
+
+## Which to use?
+
+There are some pros and cons of using `returned`, as opposed to `:=`.
+
+Firstly, `returned` is more flexible, as it allows you to track any type of object; `:=` only works with variables that can be inserted into an `MCMCChains.Chains` object.
+(Notice that `x` is a vector, and in the first case where we used `:=`, reconstructing the vector value of `x` can also be rather annoying as the chain stores each individual element of `x` separately.)
+
+However, if used carelessly, `returned` can lead to unnecessary computation.
```
mhauru, line 99: Doesn't it actually always result at least in some extra computation? Even if one uses something like `track::Bool`, there's still one extra model evaluation needed for each sample, right?

mhauru (approved):
Looks great, only a couple of optional-to-attend-to suggestions.

usage/automatic-differentiation/index.qmd
penelopeysm: I realised for Mooncake it's pretty useful to be explicit because you have to pass `config=nothing`.

usage/tracking-extra-quantities/index.qmd
penelopeysm: It does, that's true. Let me try to reword this.

penelopeysm: Okay, I changed it to focus on the extra computation during inference time – because the extra computation during the call to `returned()`, like having to run the tilde-pipeline, is usually kind of negligible, and I think the bigger problem is having some really expensive return-value being calculated on every model evaluation during sampling.

penelopeysm:
> Moving on, we generate samples from the posterior using the default NUTS sampler. We’ll make use of [ReverseDiff.jl](https://github.com/JuliaDiff/ReverseDiff.jl), as it has better performance than [ForwardDiff.jl](https://github.com/JuliaDiff/ForwardDiff.jl/) on this example. See Turing.jl’s docs on Automatic Differentiation for more info.

I just benchmarked this one, interesting - cc @willtebbutt as you might like this example where Mooncake absolutely blows RD out of the water

```julia
using DataFrames
distances = [2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20]
ns = [1443, 694, 455, 353, 272, 256, 240, 217, 200, 237, 202, 192, 174, 167, 201, 195, 191, 147, 152, ]
ys = [1346, 577, 337, 208, 149, 136, 111, 69, 67, 75, 52, 46, 54, 28, 27, 31, 33, 20, 24]
df = DataFrame(distance = distances, n = ns, y = ys)

using AbstractGPs, LogExpFunctions, Turing
@model function putting_model(d, n; jitter=1e-4)
    v ~ Gamma(2, 1)
    l ~ Gamma(4, 1)
    f = GP(v * with_lengthscale(SEKernel(), l))
    f_latent ~ f(d, jitter)
    y ~ product_distribution(Binomial.(n, logistic.(f_latent)))
    return (fx=f(d, jitter), f_latent=f_latent, y=y)
end
m = putting_model(Float64.(df.distance), df.n)

ForwardDiff 17.60233918128655
ReverseDiff 120.82069565217394
ReverseDiffCompiled 31.189760856427522
Mooncake 5.9673639441020745
```

usage/tracking-extra-quantities/index.qmd
mhauru: Looks good

Issue #583, opened 2025-02-07
Errors


penelopeysm:
One element of docs I've seen that is very helpful goes something like this:

When you get an error with a library, often the best way to figure out what went wrong is to copy paste the error message into Google.

The most helpful libraries tend to have _error codes_ which link to specific docs pages that tell you (1) why these errors happened; (2) what you can do about it; (3) possible further reading.

See e.g. the Rust compiler https://doc.rust-lang.org/error_codes/error-index.html for a stellar example of this

There is a similar effort going on with the Haskell compiler https://errors.haskell.org/

PR #603, opened 2025-05-21
Add troubleshooting page

penelopeysm:
See #583 for motivation. The idea is to have a dedicated page to the error messages that people receive and provide some guidance. It starts with only one error, but we can add to it as we go along.

Preview here: https://turinglang.org/docs/pr-previews/603/usage/troubleshooting/

I do suppose it's not mandatory to come up with error codes like `T0001` (in this case, I chose `T` because the error comes from Turing.jl; something in DynamicPPL would be `D0001`, for example).
However, I think that error codes are more stable and easier to search for than error messages (which may be changed, and may vary depending on interpolated variables).
The intent would be that after adding a code here, we can edit the message in the Turing codebase to point to this link

### DIFF
--- a/usage/troubleshooting/index.qmd
+++ b/usage/troubleshooting/index.qmd
@@ -16,6 +16,7 @@ If the suggestions here do not resolve your problem, please do feel free to [ope
 
 ```{julia}
 using Turing
+Turing.setprogress!(false)
 ```
 
 ## Initial parameters
--- a/usage/troubleshooting/index.qmd
+++ b/usage/troubleshooting/index.qmd
@@ -73,18 +73,10 @@ This example is taken from [this Turing.jl issue](https://github.com/TuringLang/
 
 ```{julia}
 @model function initial_bad2()
-	  x ~ Exponential(100)
-	  y ~ Uniform(0, x)
+    x ~ Exponential(100)
+    y ~ Uniform(0, x)
 end
 model = initial_bad2() | (y = 50.0,)
-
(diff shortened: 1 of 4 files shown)


mhauru:
I think this is a great idea. Might be worth bringing this to @yebai's attention as well, since it's an entirely new aspect of documentation. Some ideas for errors to write troubleshooting bits for:
* The error you get when you forget to do `x = Vector{T}` and just do `x = Vector{Float64}` and AD falls over.
* I thought I had a second one in mind but I don't.

penelopeysm:
@mhauru Further thoughts or approval welcome

usage/troubleshooting/index.qmd
```diff
@@ -71,11 +72,19 @@ Another cause of this error is having models with very extreme parameters.
...
 	  x ~ Exponential(100)
 	  y ~ Uniform(0, x)
 end
-model = t0001_bad2() | (y = 50.0,)
+model = initial_bad2() | (y = 50.0,)
+
+@model function initial_bad3()
+    x_trf ~ Uniform(0, 1)
+    x := -log(x_trf) / 100
+    @show x
```
mhauru, line 84: Is the `@show` used for something? Also funky indent.

mhauru:
Maybe also update the manifest to fix the version consistency check?

penelopeysm:
Ah, my fault for releasing yet another patch version.

mhauru (approved):
Thanks @penelopeysm!

PR #608, opened 2025-06-03
Add FAQ section to improve user guidance

AoifeHughes:
This pull request adds a new FAQ section to the Turing.jl website and updates the navigation menu to include a link to the FAQ page. The FAQ section provides answers to common questions about using Turing.jl, covering topics such as random variables, implementing samplers, parallelism, debugging, syntax differences, automatic differentiation backends, and performance issues.

### Website Navigation Update:
* [`_quarto.yml`](diffhunk://#diff-469a07f57393912a9c17c3ea50ceebe991e32b008beb167a200b17e91fdac6d2R27-R28): Added a new link to the FAQ page in the website's navigation menu.

### FAQ Section Addition:
* [`faq/index.qmd`](diffhunk://#diff-34eac41263e2d517abaa978de096380d28fcee5eaebdeb70af8882d09aaac4c5R1-R72): Created a new FAQ page with detailed answers to common questions about Turing.jl. Topics include variable treatment, sampler implementation, parallelism, type stability, debugging, syntax differences, automatic differentiation backends, and performance debugging.

### Relevant Issues: 

- https://github.com/TuringLang/Turing.jl/issues/2431
- https://github.com/TuringLang/Turing.jl/issues/2417
- https://github.com/TuringLang/Turing.jl/issues/2414

### DIFF
--- a/faq/index.qmd
+++ b/faq/index.qmd
@@ -1,0 +1,72 @@
+---
+title: "Frequently Asked Questions"
+description: "Common questions and answers about using Turing.jl"
+---
+
+## Why is this variable being treated as random instead of observed?
+
+This is a common source of confusion. In Turing.jl, you can only manipulate expressions that explicitly appear on the left-hand side (LHS) of a `~` statement. 
+
+For example, if your model contains:
+```julia
+x ~ filldist(Normal(), 2)
+```
+
+You cannot directly condition on `x[2]` using `condition(model, @varname(x[2]) => 1.0)` because `x[2]` never appears on the LHS of a `~` statement. Only `x` as a whole appears there.
+
+To understand more about how Turing determines whether a variable is treated as random or observed, see:
+- [Compiler Design Overview](../developers/compiler/design-overview/) - explains the heuristics Turing uses
+- [DynamicPPL Transformations](../developers/transforms/dynamicppl/) - details about variable transformations and the `@varname` macro
+- [Core Functionality](../core-functionality/) - basic explanation of the `~` notation and conditioning
+
+## How do I implement a sampler for a Turing.jl model?
+
+We have comprehensive guides on implementing custom samplers:
+- [Implementing Samplers Tutorial](../developers/inference/implementing-samplers/) - step-by-step guide on implementing samplers in the AbstractMCMC framework
+- [AbstractMCMC-Turing Interface](../developers/inference/abstractmcmc-turing/) - how to integrate your sampler with Turing
+- [AbstractMCMC Interface](../developers/inference/abstractmcmc-interface/) - the underlying interface documentation
+
+## Can I use parallelism / threads in my model?
+
+Yes! Turing.jl supports both multithreaded and distributed sampling. See the [Core Functionality guide](../core-functionality/#sampling-multiple-chains) for detailed examples showing:
+- Multithreaded sampling using `MCMCThreads()`
+- Distributed sampling using `MCMCDistributed()`
+
+## How do I check the type stability of my Turing model?
+
+Type stability is crucial for performance. Check out:
+- [Performance Tips](../usage/performance-tips/) - includes specific advice on type stability
+- [Automatic Differentiation](../usage/automatic-differentiation/) - contains benchmarking utilities using `DynamicPPL.TestUtils.AD`
+
+## How do I debug my Turing model?
+
+For debugging both statistical and syntactical issues:
+- [Troubleshooting Guide](../usage/troubleshooting/) - common errors and their solutions
+- For more advanced debugging, DynamicPPL provides `DynamicPPL.DebugUtils` for inspecting model internals
+
+## What are the main differences between Turing vs BUGS vs Stan syntax?
+
+While there are many syntactic differences, key advantages of Turing include:
+- **Julia ecosystem**: Full access to Julia's profiling and debugging tools
+- **Parallel computing**: Much easier to use distributed and parallel computing inside models
+- **Flexibility**: Can use arbitrary Julia code within models
+- **Extensibility**: Easy to implement custom distributions and samplers
+
+## Which automatic differentiation backend should I use?
+
+The choice of AD backend can significantly impact performance. See:
+- [Automatic Differentiation Guide](../usage/automatic-differentiation/) - comprehensive comparison of ForwardDiff, Mooncake, ReverseDiff, and other backends
+- [Performance Tips](../usage/performance-tips/#choose-your-ad-backend) - quick guide on choosing backends
+- [AD Backend Benchmarks](https://turinglang.org/ADTests/) - performance comparisons across various models
+
+For more specific recommendations, check out the [DifferentiationInterface.jl tutorial](https://juliadiff.org/DifferentiationInterface.jl/DifferentiationInterfaceTest/stable/tutorial/).
+
+## I changed one line of my model and now it's so much slower; why?
+
+Small changes can have big performance impacts. Common culprits include:
+- Type instability introduced by the change
+- Switching from vectorized to scalar operations (or vice versa)
+- Inadvertently causing AD backend incompatibilities
+- Breaking assumptions that allowed compiler optimizations
+
+See our [Performance Tips](../usage/performance-tips/) and [Troubleshooting Guide](../usage/troubleshooting/) for debugging performance regressions.
--- a/_quarto.yml
+++ b/_quarto.yml
@@ -24,6 +24,8 @@ website:
         text: Get Started
       - href: tutorials/coin-flipping/
         text: Tutorials
+      - href: faq/
+        text: FAQ
       - href: https://turinglang.org/library/
         text: Libraries
       - href: https://turinglang.org/news/


faq/index.qmd
```diff
@@ -0,0 +1,72 @@
+---
+title: "Frequently Asked Questions"
+description: "Common questions and answers about using Turing.jl"
+---
+
+## Why is this variable being treated as random instead of observed?
+
+This is a common source of confusion. In Turing.jl, you can only manipulate expressions that explicitly appear on the left-hand side (LHS) of a `~` statement.
```
penelopeysm, line 8:
```suggestion
This is a common source of confusion. In Turing.jl, you can only condition or fix expressions that explicitly appear on the left-hand side (LHS) of a `~` statement. 
```

faq/index.qmd
```diff
@@ -0,0 +1,72 @@
...
+
+This is a common source of confusion. In Turing.jl, you can only manipulate expressions that explicitly appear on the left-hand side (LHS) of a `~` statement. 
+
+For example, if your model contains:
+```julia
+x ~ filldist(Normal(), 2)
+```
+
+You cannot directly condition on `x[2]` using `condition(model, @varname(x[2]) => 1.0)` because `x[2]` never appears on the LHS of a `~` statement. Only `x` as a whole appears there.
+
```
penelopeysm, line 16: I think it would be good to have a bit more subtlety here, as there is a case where this isn't true: if you use `x .~ dist` where `dist` is a univariate distribution, then each element of `x` is separately treated as being drawn from `dist`. So, you can do

```julia
@model function f1()
    x = Vector{Float64}(undef, 3)
    x .~ Normal()
end

m1 = f1() | (@varname(x[1]) => 1.0)
sample(m1, NUTS(), 100) # OK
```

However, you can't condition on _part of a distribution_, so because `MvNormal()` is a distribution that returns a vector, you can't condition on part of it. (Statistically, this would be like marginalising out some degrees of freedom, but we don't have the ability to do that now.)

```julia
@model function f2()
    x = Vector{Float64}(undef, 3)
    x ~ MvNormal(zeros(3), I)
end

m2 = f2() | (@varname(x[1]) => 1.0)
sample(m2, NUTS(), 100) # Not OK
```

And `filldist` doesn't create a set of N iid distributions; it lumps them all into a single distribution, which is why you can't condition on `x[1]` in the example. So it's not so much about conditioning on the LHS of a tilde, it's more about conditioning on entire distributions at once.

It's quite confusing, and hence why I think it's important to explain this carefully.

faq/index.qmd
```diff
@@ -0,0 +1,72 @@
...
+To understand more about how Turing determines whether a variable is treated as random or observed, see:
+- [Compiler Design Overview](../developers/compiler/design-overview/) - explains the heuristics Turing uses
+- [DynamicPPL Transformations](../developers/transforms/dynamicppl/) - details about variable transformations and the `@varname` macro
```
penelopeysm, line 19:
```suggestion
```

I don't think this is relevant?

faq/index.qmd
```diff
@@ -0,0 +1,72 @@
...
+- [Core Functionality](../core-functionality/) - basic explanation of the `~` notation and conditioning
+
+## How do I implement a sampler for a Turing.jl model?
+
+We have comprehensive guides on implementing custom samplers:
+- [Implementing Samplers Tutorial](../developers/inference/implementing-samplers/) - step-by-step guide on implementing samplers in the AbstractMCMC framework
+- [AbstractMCMC-Turing Interface](../developers/inference/abstractmcmc-turing/) - how to integrate your sampler with Turing
+- [AbstractMCMC Interface](../developers/inference/abstractmcmc-interface/) - the underlying interface documentation
```
penelopeysm, line 27:
```suggestion
```

These docs are severely outdated

faq/index.qmd
```diff
@@ -0,0 +1,72 @@
...
+
+## Can I use parallelism / threads in my model?
+
+Yes! Turing.jl supports both multithreaded and distributed sampling. See the [Core Functionality guide](../core-functionality/#sampling-multiple-chains) for detailed examples showing:
+- Multithreaded sampling using `MCMCThreads()`
+- Distributed sampling using `MCMCDistributed()`
```
penelopeysm, line 33: This one also needs more nuance. Sampling one chain per thread is indeed done with `MCMCThreads()` but (I think) the main intent behind this question is whether one can use threading _within the model itself_, e.g.,

```julia
@model function f(x)
    Threads.@threads for i in eachindex(x)
        x[i] ~ Normal()
    end
end
```

That means that the execution of the model itself is inherently multithreaded (as opposed to sampling with MCMCThreads(), where each chain runs the model many times but only on one thread).

The TLDR of this in the model is that threaded observe statements are OK but threaded assume statements are not (they often crash in unpredictable ways, or sometimes they actually work if you're really lucky).

Another aspect maybe worth mentioning is that threads inside models don't work with many AD backends (see `multithreaded` in https://turinglang.org/ADTests/ for an example)

faq/index.qmd
```diff
@@ -0,0 +1,72 @@
...
+
+## How do I check the type stability of my Turing model?
+
+Type stability is crucial for performance. Check out:
+- [Performance Tips](../usage/performance-tips/) - includes specific advice on type stability
+- [Automatic Differentiation](../usage/automatic-differentiation/) - contains benchmarking utilities using `DynamicPPL.TestUtils.AD`
```
penelopeysm, line 39:
```suggestion
```
The AD page doesn't discuss type stability.

I think the best 'high-level' way to determine type stability is https://turinglang.org/DynamicPPL.jl/stable/api/#DynamicPPL.DebugUtils.model_warntype although I've not used it myself.

faq/index.qmd
```diff
@@ -0,0 +1,72 @@
...
+- [Troubleshooting Guide](../usage/troubleshooting/) - common errors and their solutions
+- For more advanced debugging, DynamicPPL provides `DynamicPPL.DebugUtils` for inspecting model internals
+
+## What are the main differences between Turing vs BUGS vs Stan syntax?
+
+While there are many syntactic differences, key advantages of Turing include:
+- **Julia ecosystem**: Full access to Julia's profiling and debugging tools
+- **Parallel computing**: Much easier to use distributed and parallel computing inside models
+- **Flexibility**: Can use arbitrary Julia code within models
+- **Extensibility**: Easy to implement custom distributions and samplers
```
penelopeysm, line 53: This doesn't answer the stated question?

If it's a question about syntax, I would personally think it's best to have Turing vs Stan in a single section, and Turing vs (other PPL) can be a different section. Apart from the parameter block, two of the common things that Stan models have that Turing doesn't are `transformed data` and `generated quantities` -- the gist in Turing is that any data transformations should be done before defining the model (so it's decoupled from the model) and generated quantities are covered in https://turinglang.org/docs/usage/tracking-extra-quantities/index.html.

faq/index.qmd
```diff
@@ -0,0 +1,72 @@
...
+
+## Which automatic differentiation backend should I use?
+
+The choice of AD backend can significantly impact performance. See:
+- [Automatic Differentiation Guide](../usage/automatic-differentiation/) - comprehensive comparison of ForwardDiff, Mooncake, ReverseDiff, and other backends
+- [Performance Tips](../usage/performance-tips/#choose-your-ad-backend) - quick guide on choosing backends
+- [AD Backend Benchmarks](https://turinglang.org/ADTests/) - performance comparisons across various models
+
+For more specific recommendations, check out the [DifferentiationInterface.jl tutorial](https://juliadiff.org/DifferentiationInterface.jl/DifferentiationInterfaceTest/stable/tutorial/).
```
penelopeysm, line 62: I don't think DITest is relevant yet --- to use the benchmarking functionality in there we would need to have a way of constructing `DifferentiationInterfaceTest.Scenario`s from Turing models and although it's not very hard to write some code for that, it's probably too long to go into an FAQ page. Would be a good thing to add to the AD page, probably in a separate PR.

faq/index.qmd
```diff
@@ -0,0 +1,72 @@
...
+
+## I changed one line of my model and now it's so much slower; why?
+
+Small changes can have big performance impacts. Common culprits include:
+- Type instability introduced by the change
+- Switching from vectorized to scalar operations (or vice versa)
+- Inadvertently causing AD backend incompatibilities
+- Breaking assumptions that allowed compiler optimizations
+
+See our [Performance Tips](../usage/performance-tips/) and [Troubleshooting Guide](../usage/troubleshooting/) for debugging performance regressions.
```
penelopeysm, line 72: I know you took the list questions from somewhere else, but I don't really like this question. I don't get the intent behind it (what is the answer supposed to be?) and the result of this is, I think, reflected in the text, which is very vague and IMO not very helpful. If the answer is basically to read the performance section.

IME interactions with AD backend don't often lead to performance differences, usually it either runs fine or it crashes. If the AD is unusually slow it usually reflects slowness in the model itself.

penelopeysm (changes requested):
This is a nice start, I think there are a few places where it's necessary to go into more detail, which I've commented on.

General comment for the entire PR: for the URLs, you can use the meta variables defined in `_quarto.yml`, i.e. `[text]({{< meta usage-automatic-differentiation >}})` to avoid hardcoding relative paths (useful if you want to move pages around, which has happened a couple of times).

faq/index.qmd
penelopeysm: Sure, there's a pinned issue with a huge list of suggested topics to cover with docs -- enough to keep any of us occupied for a year 😅

penelopeysm: I went through and the only ones that are unforgivably outdated (to the point where I would suggest a rewrite is better) are:

- `developers/inference/abstractmcmc-interface`
- `developers/inference/abstractmcmc-turing`
- `developers/compiler/design-overview`

faq/index.qmd
```diff
@@ -0,0 +1,137 @@
...
+
+### 2. Threading Within Models
+Using threads inside your model (e.g., `Threads.@threads`) requires more care:
+
+```julia
+@model function f(x)
+    Threads.@threads for i in eachindex(x)
+        x[i] ~ Normal()  # UNSAFE: Assume statements in threads can crash!
+    end
+end
```
penelopeysm, line 66: In this model, `x` is observed so it's fine. 'assume' means that it's a random parameter, so for example something like this:

```suggestion
@model function f(y)
    x = Vector{Float64}(undef, length(y))
    Threads.@threads for i in eachindex(y)
        x[i] ~ Normal()  # UNSAFE: `assume` statements in @threads can crash!
        y[i] ~ Normal(x[i])  # Observe statements are okay
    end
end
```

faq/index.qmd
```diff
@@ -0,0 +1,137 @@
...
+
+Type stability is crucial for performance. Check out:
+- [Performance Tips]({{< meta usage-performance-tips >}}) - includes specific advice on type stability
+- Use `DynamicPPL.DebugUtils.model_warntype` to check type stability of your model
+
+## How do I debug my Turing model?
+
+For debugging both statistical and syntactical issues:
+- [Troubleshooting Guide]({{< meta usage-troubleshooting >}}) - common errors and their solutions
+- For more advanced debugging, DynamicPPL provides `DynamicPPL.DebugUtils` for inspecting model internals
```
penelopeysm, line 86:
```suggestion
- For more advanced debugging, DynamicPPL provides [the `DynamicPPL.DebugUtils` module](https://turinglang.org/DynamicPPL.jl/stable/api/#Debugging-Utilities) for inspecting model internals
```

Friendly link

penelopeysm (changes requested):
Looking good! Thanks for expanding on the points here. A few more comments, which hopefully aren't too complicated to deal with.

faq/index.qmd
```diff
@@ -0,0 +1,141 @@
...
+- **Transformed data**: Stan has a `transformed data` block for preprocessing. In Turing, data transformations should be done before defining the model
+- **Generated quantities**: Stan has a `generated quantities` block. In Turing, use the approach described in [Tracking Extra Quantities]({{< meta usage-tracking-extra-quantities >}})
+
+Example comparison:
+```stan
+// Stan
+data {
+  int<lower=0> N;
+  vector[N] y;
+}
```
penelopeysm, line 105: Sorry I didn't catch this last time, but I think to be unambiguously consistent with the Turing model, `y` should be a scalar.

(To be technically precise, the Turing model will work if you give it a vector `y`, but that's because of some implicit broadcasting that both Markus and I have complained about before: https://github.com/TuringLang/DynamicPPL.jl/issues/919)

penelopeysm (commented):
The build failure isn't related to this PR, it's because MicroCanonicalHMC.jl has rather questionable behaviour. I'll fix it separately and merge into this branch.

penelopeysm:
Did you see my previous comment (and the other PR)?

penelopeysm:
The build failure is because of a different package, not this PR, and I fixed it in a separate PR, and pinged you to review - please scroll up a bit

penelopeysm (approved):

yebai:
@shravanngoswamii can you add the FAQ nav menu everywhere?

shravanngoswamii:
> @shravanngoswamii can you add the FAQ nav menu everywhere?

Sure!

PR #613, opened 2025-06-09
Color-Theme update to match main site for consistency

shravanngoswamii:
It should be merged after https://github.com/TuringLang/turinglang.github.io/pull/119


simonsteiger:
Something unexpected came up yesterday afternoon and I wasn't able to spend much time with this PR as a result. I think that the theming is good to go, and I'm happy to play with the code blocks a bit more when I find the time.

PR #618, opened 2025-06-17
Fix AD page, add ForwardDiff troubleshooting thing

penelopeysm:
#595, apart from attempting to define 'support tiers' for AD backends, also revamped most of the text on that page.

We couldn't quite agree on what support tiers were, so I abandoned that PR, but I think the rest of the content is still worth putting here.

In #595 I also wrote a section describing the problem associated with ForwardDiff Duals and vectors of Float. I've thus also added this to the new troubleshooting page (as was suggested in https://github.com/TuringLang/docs/pull/603#issuecomment-2901097739). I addressed Will's comment https://github.com/TuringLang/docs/pull/595#discussion_r2028394085 about making it clear that the type parameter option is better than making a `Vector{Real}`.

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -2,9 +2,6 @@ name: PR Preview Workflow
 
 on:
   pull_request:
-    types:
-      - opened
-      - synchronize
 
 concurrency:
   group: docs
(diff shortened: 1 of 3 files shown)


penelopeysm:
None of the text here is new and we already had two pairs of eyes on it in #595, so I think I'm just going to merge once CI passes.

penelopeysm:
grrrrr

Issue #576, opened 2025-01-16
Remove Pathfinder code example from docs


penelopeysm:
Pathfinder.jl is lovely, but we probably shouldn't have a code example in there:

https://github.com/TuringLang/docs/blob/2260d618fdca9073f05c7e25c3c078aeb256b1c7/tutorials/docs-16-using-turing-external-samplers/index.qmd#L82-L99

Rationale:

1. We want the docs to use the most recent version of Turing.jl, as far as is possible. Pathfinder is the only reverse dep of Turing that we are using*, meaning that if we release a new version of Turing, we need Pathfinder to also release a new version that is compatible with it. We shouldn't expect Pathfinder's maintainers to do this (they may not want to, for whatever reason).

2. We shouldn't be responsible for documenting how another package is used. It's great to point out other packages in the Julia ecosystem that can be used together with Turing, but the question of how to use package X should be documented in package X. Indeed, Pathfinder's docs already contain an example that is remarkably similar to our own: https://mlcolab.github.io/Pathfinder.jl/stable/examples/turing/

\* There's also TuringBenchmarking, but that's internal, so we can easily tag a new release whenever we need to.

penelopeysm:
Closed by #579

PR #628, opened 2025-07-31
Remove MicroCanonicalHMC.jl and update external sampler docs

penelopeysm:
Several reasons:

1. MCHMC.jl is a reverse dependency of Turing (i.e., it depends on Turing). We previously agreed that it's not generally desirable to have reverse-deps in the docs repo (#576). The TL;DR is that reverse deps prevent us from releasing docs with new versions of Turing until and unless those reverse deps upgrade their Turing compat entries, and we can't control if or when that happens.

2. MCHMC.jl hasn't been updated in quite a while, in fact its latest release declares a Turing compat version of 0.34 which is almost a year old https://github.com/TuringLang/Turing.jl/releases/tag/v0.34.1.

3. Well, in that case how have the docs even been building for the last 5 minor releases of Turing? Turns out that old versions of MCHMC.jl have no compat bounds. So we have just been building with an old release of MCHMC that fortuitously didn't error, but which should really be yanked from the General registry. I've opened an issue on the MCHMC repo about this: https://github.com/JaimeRZP/MicroCanonicalHMC.jl/issues/27.

In its place I have put a link to SliceSampling.jl (which avoids problems (2) and (3)) but without any actual code (which avoids problem (1)).

In this PR I also took the opportunity to completely refresh the interface documentation for external samplers, which was quite outdated.


penelopeysm:
@AoifeHughes, if you merge this into main and merge main into your PR, then CI should pass. Of course, please do feel free to review this PR as usual before merging and ping me if there is anything that isn't clear.

penelopeysm:
Arghh, still not fixed. It's a Turing bug.

penelopeysm:
OK, should be fixed now and ready for review.

PR #629, opened 2025-08-04
Add documentation for Stochastic Gradient Samplers

AoifeHughes:
- https://github.com/TuringLang/Turing.jl/issues/2270 adds docs to support this

### DIFF
--- a/usage/stochastic-gradient-samplers/index.qmd
+++ b/usage/stochastic-gradient-samplers/index.qmd
@@ -114,7 +114,9 @@ plot(chain_sghmc[5001:end])
 For comparison, let's sample the same model using standard HMC:
 
 ```{julia}
-chain_hmc = sample(model, HMC(0.01, 10), 1000)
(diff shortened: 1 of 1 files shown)


mhauru:
I don't know why HMC is having such trouble with this quite simple model, the numerical integration errors just blow up quite often. Seems that you can fix that with a decent initial value for the chain though. Try adding the keyword argument `initial_params=[0.0, 1.0]` to the `sample` call for HMC, that seems to help.

We should set the same `initial_params` for the two other samplers as well. Fixing the starting point of the MCMC chain also makes the comparison between the different samplers a bit fairer, so not a bad thing in general. However, it doesn't solve all the convergence issues with at least SGLD (I didn't look at the other one). However, those seem to be fixable by tuning the parameters of `PolynomialStepsize`. I don't really understand `PolynomialStepsize`, but at least for SGLD `PolynomialStepsize(0.01, 100)` seems to work decently. Try adding in that, and if necessary tune the parameters for the other sampler as well.

PR #632, opened 2025-08-07
Added url redirects

shravanngoswamii:
We have to keep atleast two redirect files in this directory -- Quarto lisiting behavior, we have to deal with it!

Ref: https://github.com/TuringLang/Turing.jl/pull/2637#discussion_r2259805866

### DIFF
--- a/uri/_metadata.yml
+++ b/uri/_metadata.yml
@@ -1,0 +1,2 @@
+listing: 
+    template: redirect.ejs
--- a/uri/initial-parameters.qmd
+++ b/uri/initial-parameters.qmd
@@ -1,0 +1,5 @@
+---
+title: "Troubleshooting - Initial parameters"
+aliases: [initial-parameters]
+redirect: https://turinglang.org/docs/usage/troubleshooting/#initial-parameters
+---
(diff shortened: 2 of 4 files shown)


shravanngoswamii:
https://turinglang.org/docs/pr-previews/632/uri/initial-parameters redirects to https://turinglang.org/docs/usage/troubleshooting/#initial-parameters

yebai:
> Would it be sufficient to add an aliases in the troubleshooting page itself? And maybe add a comment in the page itself to say that 'this URL is used in Turing.jl here, if you move this file please make sure to retain it'?

I like @penelopeysm's proposal here. The current approach seems overkill -- it's unlikely we will have lots of redirects.

shravanngoswamii:
I agree to both of you but using `aliases` in file itself will redirect to `https://turinglang.org/docs/usage/troubleshooting/` page and not some particular section link like https://turinglang.org/docs/usage/troubleshooting/#initial-parameters -- We can only achieve it using current approach.

shravanngoswamii:
@AoifeHughes Please feel free to merge this based on your requirements in this PR -- https://github.com/TuringLang/Turing.jl/pull/2637#discussion_r2259805866

PR #636, opened 2025-08-13
Update to Turing 0.40

mhauru:


penelopeysm:
I wonder if it's worth writing something in the dev docs about contexts vs accumulators and when one should use which?

developers/compiler/minituring-contexts/index.qmd
```diff
@@ -14,6 +14,8 @@ Pkg.instantiate();
 
 In the [Mini Turing]({{< meta minituring >}}) tutorial we developed a miniature version of the Turing language, to illustrate its core design. A passing mention was made of contexts. In this tutorial we develop that aspect of our mini Turing language further to demonstrate how and why contexts are an important part of Turing's design.
 
+Note: The way Turing actually uses contexts changed somewhat in releases 0.39 and 0.40. The content of this page remains relevant, the principles of how contexts operate remain the same, and concepts like leaf and parent contexts still exist. However, we've moved away from using contexts for quite as many things as we used to. Most importantly, whether to accumulate the log joint, log prior, or log likelihood is no longer done using different contexts. Please keep this in mind as you read this page: The principles remain, but the details have changed. We will update this page once the refactoring of internals that is happening around releases like 0.39 and 0.40 is done.
```
penelopeysm, line 17:
```suggestion
::: {.callout-important}
Note: The way Turing actually uses contexts changed somewhat in releases 0.39 and 0.40. The content of this page remains relevant, the principles of how contexts operate remain the same, and concepts like leaf and parent contexts still exist. However, we've moved away from using contexts for quite as many things as we used to. Most importantly, whether to accumulate the log joint, log prior, or log likelihood is no longer done using different contexts. Please keep this in mind as you read this page: The principles remain, but the details have changed. We will update this page once the refactoring of internals that is happening around releases like 0.39 and 0.40 is done.
:::
```

https://quarto.org/docs/authoring/callouts.html

usage/modifying-logprob/index.qmd
```diff
@@ -47,13 +47,10 @@ using LinearAlgebra
 end
 ```
 
-Note that `@addlogprob!` always increases the accumulated log probability, regardless of the provided
-sampling context.
-For instance, if you do not want to apply `@addlogprob!` when evaluating the prior of your model but only when computing the log likelihood and the log joint probability, then you should [check the type of the internal variable `__context_`](https://github.com/TuringLang/DynamicPPL.jl/issues/154), as in the following example:
+Note that `@addlogprob!` increases the accumulated log likelihood.
```
penelopeysm, line 50:
```suggestion
Note that `@addlogprob! (p::Float64)` adds `p` to the log likelihood.
```

penelopeysm (changes requested):
Some minor suggestions for wording / formatting.

While it's already quite useful getting the docs to build with a version of Turing, I have a nagging feeling that we are not documenting the things that we should be documenting very well.

For one, it looks like documenting the whole accumulators / contexts thing is just being shunted until we have more stability around the interface. That's at least sort of understandable, because who knows what's going to change in upcoming releases.

However, there are also the user-facing changes that arise from that. For example, accumulators mean that chains now contain logp + its components. This is obviously worth showing somewhere, but it isn't being shown.

I don't want that to hold this PR up, but I do think that at least new issue(s) should be made for anything that needs to be added to the docs.

mhauru:
I agree that we should add/improve the content. I opened these issues to track:
* https://github.com/TuringLang/docs/issues/646
* https://github.com/TuringLang/docs/issues/644
* https://github.com/TuringLang/docs/issues/645

My goal with this PR would be to get the code examples working, and to hopefully not state anything that is plain wrong, even if it falls short of explaining all that is there.

PR #640, opened 2025-08-21
Make light-mode teal colour darker

### LINKED ISSUE #639: The teal colour in light mode is too low-contrast
In general I like the teal colour a lot, but it has been straining my eyes a bit. Would you be OK with making it a bit darker @shravanngoswamii? I recognise the main site will need to be updated in tandem too.

Before:

Image

After:

Image

penelopeysm:
Closes #639.

### DIFF
--- a/theming/variables/_colors-light.scss
+++ b/theming/variables/_colors-light.scss
@@ -1,7 +1,7 @@
 @import "grays";
 
 $primary: #2780e3;
-$teal-link: #34b8bf;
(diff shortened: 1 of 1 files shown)


penelopeysm:
I did take a look at the dark mode and I thought it was fine. I suppose it's because it's against a dark background so it's fine -- the thing my eyes were struggling with was a fairly light blue against a fairly light grey :)

I don't use dark mode myself though so I might not be the best judge 👀

PR #642, opened 2025-09-18
Add download button for Jupyter notebooks in tutorial pages

AoifeHughes:
No need to review yet, I first want to test and verity the deployment works properly


_quarto.yml
```diff
@@ -153,6 +153,11 @@ format:
     toc-title: "Table of Contents"
     code-fold: false
     code-overflow: scroll
+    code-tools:
+      source: true
+      toggle: false
+      caption: none
+    downloads: [ipynb]
```
shravanngoswamii, line 160: I think this `downloads` option only works for books, not sure!

shravanngoswamii:
I think we are looking for this: https://quarto.org/docs/output-formats/html-multi-format.html#overview

shravanngoswamii:
> I don't need reviewers on this at the moment. Thanks!
Hey, sorry about the unsolicited review and force push — I was going through the docs after a long time and ended up poking around your PR out of habit. Definitely didn’t mean to interfere.

PR #643, opened 2025-09-18
Update footer text to specify Turing.jl

AoifeHughes:
Changed `Turing` -> `Turing.jl` in footer.


penelopeysm (approved):
> lemme just hit the update button for a bunch of packages because CI demands it

CI doesn't demand it -- it's not all the other package versions that matter, it's only the version of Turing that matters. So unless you are specifically updating the Turing version, like what is happening in #636, you don't need to update deps.

It also saves CI time to not update the non-Turing deps because then Quarto can reuse a lot of cached things.

PR #649, opened 2025-09-26
Fix list of important limitations in the FAQ

seabbs:
On the site I see: 

Screenshot 2025-09-26 at 14 39 18

and similar. Adding a space should fix this which I have done. 

A few/all of these maybe intentional but I think that this is not the case. Happy to revert some as directed or happy for this to be closed.


penelopeysm:
Thanks! That's indeed better :)

PR #667, opened 2025-11-24
Threadsafe evaluation

### LINKED ISSUE #656: Within-model parallelism.
I wrote up a long explanation here: https://discourse.julialang.org/t/parallelism-within-turing-jl-model/54064/7?u=penelopeysm

This should definitely be added to the docs

penelopeysm:
Closes #656.

Preview here: https://turinglang.org/docs/pr-previews/667/usage/threadsafe-evaluation/

I would have liked to add a TSVI vs no-TSVI benchmark to demonstrate the performance difference, but the problem is that that will require us to release DPPL 0.39 + a compatible Turing version. So at this point in time, it is not possible to write that on this repo.

### DIFF
--- a/usage/threadsafe-evaluation/index.qmd
+++ b/usage/threadsafe-evaluation/index.qmd
@@ -258,7 +258,7 @@ As it happens, much of what is needed in DynamicPPL can be constructed such that
 For example, as long as there is no need to *sample* new values of random variables, it is actually fine to completely omit the metadata object.
 This is the case for `LogDensityFunction`: since values are provided as the input vector, there is no need to store it in metadata.
 We need only calculate the associated log-prior probability, which is stored in an accumulator.
-Thus, starting from DynamicPPL v0.39, `LogDensityFunction` itself is in fact completely threadsafe.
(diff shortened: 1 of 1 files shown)


usage/threadsafe-evaluation/index.qmd
```diff
@@ -0,0 +1,270 @@
...
+However, `assume` tilde-statements always modify the metadata, and thus cannot currently be made threadsafe.
+
+### OnlyAccsVarInfo
+
+As it happens, much of what is needed in DynamicPPL can be constructed such that they *only* rely on accumulators.
+
+For example, as long as there is no need to *sample* new values of random variables, it is actually fine to completely omit the metadata object.
+This is the case for `LogDensityFunction`: since values are provided as the input vector, there is no need to store it in metadata.
+We need only calculate the associated log-prior probability, which is stored in an accumulator.
+Thus, starting from DynamicPPL v0.39, `LogDensityFunction` itself is in fact completely threadsafe.
```
mhauru:
```suggestion
Thus, starting from DynamicPPL v0.39, `LogDensityFunction` itself will in fact be completely threadsafe.
```

penelopeysm:
Good catch! 😄

PR #682, opened 2026-03-05
Update for v0.43

penelopeysm:
Replaced by #683

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -4,7 +4,8 @@ on:
   pull_request:
 
 concurrency:
-  group: docs
+  group: ${{ github.workflow }}-${{ github.ref || github.run_id }}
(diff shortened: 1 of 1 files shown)


penelopeysm:
Precompilation tends to fail, and it's impossible to update things to the latest versions. The single big environment worked so well for the last 1.5 years, but something is going wrong. I'm trying to remove deps that we don't need to unblock things, but it's tricky.

penelopeysm:
On the bright side, removing DifferentialEquations.jl (in favour of ODE.jl and DDE.jl), and replacing the [now deprecated MLDataUtils](https://github.com/JuliaML/MLDataUtils.jl?tab=readme-ov-file) with MLUtils, seems to have unblocked the latest versions of Mooncake + SciMLSens. So maybe the issue was just stale deps.

penelopeysm:
See #683

PR #690, opened 2026-04-22
Update for Turing v0.44

penelopeysm:
~~I expect the VI tutorial will fail (I need to fix it offline). However, I'm pushing it now so that CI can get started on its stuff.~~ VI tutorial fixed now, so just need to see what else breaks.


penelopeysm:
https://github.com/SciML/SciMLBase.jl/issues/1325 broke the DE page

PR #697, opened 2026-05-02
Turing v0.45

### LINKED ISSUE #483: Better describe how to get data out of `chain` elements in tutorials
In the tutorial https://turinglang.org/v0.30/tutorials/10-bayesian-differential-equations/ , there is very little detail on how to obtain the parameters shown in the summary statistics (so that you may end up being stuck with a pretty print but no obvious way to access the information that is printed)
```raw

Summary Statistics
  parameters      mean       std      mcse   ess_bulk   ess_tail      rhat 
  e ⋯
      Symbol   Float64   Float64   Float64    Float64    Float64   Float64 
    ⋯

           σ    0.8440    0.0640    0.0086    51.8543    50.1666    1.0926 
    ⋯
           α    1.5362    0.1946    0.0292    44.8038   115.2958    1.3519 
    ⋯
           β    1.0710    0.1563    0.0229    46.9437   158.4708    1.3206 
    ⋯
           γ    3.0231    0.2973    0.0438    48.2945   132.1955    1.3178 
    ⋯
           δ    0.9911    0.2736    0.0420    45.0081   125.5487    1.3476 
    ⋯
                                                                1 column om
itted

Quantiles
  parameters      2.5%     25.0%     50.0%     75.0%     97.5%
      Symbol   Float64   Float64   Float64   Float64   Float64

           σ    0.7221    0.8021    0.8416    0.8840    0.9851
           α    1.2433    1.3775    1.5346    1.6588    1.9596
           β    0.8278    0.9431    1.0582    1.1714    1.4067
           γ    2.4578    2.8127    3.0105    3.2702    3.5290
           δ    0.5137    0.7841    0.9607    1.2019    

- [ ] 1.5055

```

I think one should add that after that line of code
```jl
# Sample 3 independent chains.
chain2 = sample(model2, NUTS(0.45), MCMCSerial(), 5000, 3; progress=false)
```
that one can obtain these tables by calling
```jl
describe(chain2)
```
and/or
```jl
summarystats(chain2)
```
That you can estimate the mean parameters in the table with
```jl
mean(chain2[:α])
```
and that you can run
```jl
rhat(chain2)
```
to obtain the `rhat` metric. 

This also points to a larger issue : This page https://turinglang.org/MCMCChains.jl/stable/diagnostics/#MCMCDiagnosticTools.rhat-Tuple{Chains} is not searchable from https://turinglang.org so that any user searching `rhat` in the search bar would find no information whatsoever.

The same is true for `describe` and `summarystats`, found in another page of the MCMCCChain documentation https://turinglang.org/MCMCChains.jl/stable/stats/#StatsBase.summarystats but not searchable from the turinglang.org website.

Although the MCMCChain documentation page is part of the website (go to Library API > Diagnostics > MCMCChains) it is not included in search results from turinglang.org, so one would be forgiven for thinking that there was no documentation on these functions when using the website.

So ideally what I would propose is 

- [ ] Make search on https://turinglang.org/  global so that it also searches inside all of the libraries that are part of the Library API section of the website)
- [ ] Add clear instructions on tutorial pages on how to extract all (or some of) the information provided in the result tables (including `describe(chain2)` in the tutorial code and link to the MCMC diagnostics page for finer control would be enough for user to find all of the needed information they need on their own)

### LINKED ISSUE #677: Remove minituring pages
Sorry to all those whose effort went into this, but it's now rather outmoded, and the way we overload assume directly in a sampler is also no longer the recommended workflow.

penelopeysm:
These are all just interface changes for MCMCChains -> FlexiChains, there aren't any other content changes. I also took the liberty to remove the minituring pages as they're several versions outdated (we don't have `SamplingContext` or `assume` anymore).

Closes #677
Inadvertently closes #483 since FlexiChains has better docs


usage/sampling-options/index.qmd
```diff
@@ -232,7 +230,8 @@ Finally, note that the first sample in the resumed chain will not be the same as
 ```{julia}
 # In general these will not be the same (although it _could_ be if the MH step
 # was rejected -- that is why we seed the sampling in this section).
-chn1[:x][end], chn2[:x][1]
+
+chn1[@varname(x)][end], chn1[@varname(y)][end]
```
sunxd3: I'm a little confused, should this be comparing chn1[y] with chn2[x]?

usage/predictive-distributions/index.qmd
```diff
@@ -97,16 +97,16 @@ This controls the generation of new `X` samples, and makes your results reproduc
 :::
 
 ::: {.callout-note}
-`predict` returns a Chains object itself, which will only contain the newly predicted variables.
-If you want to also retain the original parameters, you can use `predict(rng, predictive_model, chain; include_all=true)`.
+`predict` returns a FlexiChain object itself, which will contain all the original parameters plus the newly predicted variables.
+If you want only include the newly predicted variables, you can use `predict(rng, predictive_model, chain; include_all=false)`.
```
sunxd3:
```suggestion
If you want to only include the newly predicted variables, you can use `predict(rng, predictive_model, chain; include_all=false)`.
```

tutorials/infinite-mixture-models/index.qmd
```diff
@@ -260,7 +260,7 @@ Finally, we can plot the number of clusters in each sample.
 ```{julia}
 # Extract the number of clusters for each sample of the Markov chain.
 k = map(
-    t -> length(unique(vec(chain[t, MCMCChains.namesingroup(chain, :z), :].value))),
+    t -> length(unique(chain[@varname(z), iter=t])),
```
sunxd3: (courtesy of Claude)

`chain[@varname(z), iter=t]` keeps the chain dimension, so for a single-chain run with N-element `z` it returns a `1×N` `DimMatrix`. `unique` on a 2-D array returns unique *rows*, not unique elements:

```julia
unique([1 2 3])  # → [[1 2 3]] — length 1
```

So `k` will end up `1` for nearly every iteration and the cluster-count plot won't match the prose right below it about converging to ~3 clusters.

```suggestion
    t -> length(unique(chain[@varname(z), iter=t, chain=1])),
```

sunxd:
https://turinglang.org/docs/pr-previews/697/tutorials/infinite-mixture-models/ near the bottom the plot does look wrong

tutorials/hidden-markov-models/index.qmd
```diff
@@ -157,19 +174,14 @@ Looks like our model did a pretty good job, but we should also check to make sur
...
 A cursory examination of the traceplot above indicates that all three chains converged to something resembling
-stationary. We can use the diagnostic functions provided by [MCMCChains](https://github.com/TuringLang/MCMCChains.jl) to engage in some more formal tests, like the Heidelberg and Welch diagnostic:
-
-```{julia}
-heideldiag(MCMCChains.group(chn, :T))[1]
-```
-
-The p-values on the test suggest that we cannot reject the hypothesis that the observed sequence comes from a stationary distribution, so we can be reasonably confident that our transition matrix has converged to something reasonable.
+stationary.
+We can use the diagnostic functions provided by [MCMCDiagnosticTools](https://github.com/TuringLang/MCMCDiagnosticTools.jl) to engage in some more formal tests, like the Heidelberg and Welch diagnostic.
```
sunxd3:
```suggestion
We can use the diagnostic functions provided by [MCMCDiagnosticTools](https://github.com/TuringLang/MCMCDiagnosticTools.jl) to engage in some more formal tests, like the [Heidelberg and Welch diagnostic](https://turinglang.org/MCMCDiagnosticTools.jl/stable/#MCMCDiagnosticTools.heideldiag).
```

usage/sampling-options/index.qmd
```diff
@@ -183,30 +183,28 @@ typeof(saved_state)
...
 ```
 
-Note that the exact format saved in `chn.info.samplerstate`, and that expected by `initial_state`, depends on the invocation of `sample` used.
-For single-chain sampling, the saved state, and the required initial state, is just a single sampler state.
-For multiple-chain sampling, it is a vector of states, one per chain.
-
 ::: {.callout-warning}
 ## `resume_from`
 
 The `resume_from` argument has been removed in Turing v0.41; please use `initial_state=loadstate(chn)` instead, as described here.
```
sunxd3: sometimes it's `only(loadstate(chn))`?

sunxd3 (commented):
thanks for doing this, Penny

usage/sampling-options/index.qmd
penelopeysm: Oh, yeah, somehow I got confused myself.

tutorials/infinite-mixture-models/index.qmd
penelopeysm: I had no idea that `unique` did that and must have missed this one, thanks

tutorials/hidden-markov-models/index.qmd
penelopeysm: Thanks, updated. I haven't implemented heideldiag for FlexiChains yet sadly

usage/sampling-options/index.qmd
penelopeysm: I guess it depends on the exact invocation of `sample` so it's a bit finicky, but I rephrased to clarify.

tutorials/infinite-mixture-models/index.qmd
penelopeysm: So Claude's not right:

```julia
julia> unique([1 2 3])
3-element Vector{Int64}:
 1
 2
 3
```

But I realised what the issue was, it's that `chain[@varname(z)]` is an array of vectors, not a 3D matrix. So just passing `stack=true` makes that right.

sunxd3: nice! Sorry I should have verified

sunxd3 (approved):

penelopeysm:
Thank you! :)

Issue #700, opened 2026-07-16
Turing performance, debugging and benchmarks meta issue


shravanngoswamii:
Tracking issue for documentation on measuring and improving the performance of Turing models, and on debugging them. It collects the following:

- TuringLang/docs#702 (better performance guidelines: benchmark and profile a model, choose an AD backend, write efficient models)
- TuringLang/Turing.jl#564 (a page on debugging tools)
- TuringLang/Turing.jl#565 (tips on optimising performance)
- TuringLang/Turing.jl#668 (interpreting chain summary statistics and debugging MCMC)

### What the docs cover today

- Choosing an AD backend: [Automatic Differentiation](https://turinglang.org/docs/usage/automatic-differentiation/) (heuristics, `run_ad` benchmarking) and a FAQ entry.
- Writing efficient models: [Performance Tips](https://turinglang.org/docs/usage/performance-tips/) (multivariate distributions, AD backend choice, type stability).
- Error-level debugging: [Troubleshooting](https://turinglang.org/docs/usage/troubleshooting/).

Note that TuringBenchmarking.jl is deprecated. The current tool is `DynamicPPL.TestUtils.AD.run_ad`, which gives primal and gradient timings per backend.

### Remaining work

- [x] Benchmarking and profiling guidance (Turing.jl#2414, TuringLang/Turing.jl#565)
- [x] A page on model debugging tools: `check_model`, `model_warntype`, and related utilities (#564)
- [x] Remove references to the archived ADTests site
- [ ] A guide on interpreting chain summary statistics (`ess`, `rhat`, `mcse`) and diagnosing MCMC problems (#668)

The first three items are addressed by the linked PR. The last is a larger writing task and is left open here.

PR #701, opened 2026-07-16
Add debugging page and benchmarking and profiling guidance

### LINKED ISSUE #564: Add a page on debugging tools for Turing models
See, e.g., https://turinglang.org/DynamicPPL.jl/stable/api/#Debugging-Utilities

### LINKED ISSUE #565: Add tips on optimising performance for Turing models
~~Right now our only mention of TuringBenchmarking is in the AD bit~~

shravanngoswamii:
Part of #700.

- Adds a Debugging Models page. Closes #564.
- Adds a benchmarking and profiling section to the Performance Tips page. Closes #565 and addresses the documentation part of TuringLang/Turing.jl#2414.
- Removes all references to ADTests, which is now archived.

### DIFF
--- a/usage/debugging/index.qmd
+++ b/usage/debugging/index.qmd
@@ -1,0 +1,74 @@
+---
+title: Debugging Models
+engine: julia
+---
+
+```{julia}
+#| echo: false
+#| output: false
+using Pkg;
+Pkg.instantiate();
+```
+
+This page describes the tools available for checking that a Turing model is defined correctly.
+It complements the [Troubleshooting]({{<meta usage-troubleshooting>}}) page, which explains specific error messages: the tools here help you find problems before (or without) hitting an error.
+
+```{julia}
+using Turing
+```
+
+## Checking model validity with `check_model`
+
+`DynamicPPL.check_model` evaluates a model once and checks it for common problems, such as the same variable appearing on the left-hand side of more than one tilde statement, or `NaN` values being observed.
+It returns `true` if the check passes, and `false` (with warnings explaining the problem) otherwise.
+
+For example, this model observes a `NaN` value, which usually indicates a bug in the data preparation:
+
+```{julia}
+using DynamicPPL: check_model
+
+@model function gaussian(y)
+    x ~ Normal()
+    y ~ Normal(x)
+end
+
+check_model(gaussian(NaN))
+```
+
+whereas the same model with valid data passes:
+
+```{julia}
+check_model(gaussian(1.5))
+```
+
+Turing runs this check automatically at the start of `sample` (and the mode estimation functions).
+If the check reports a false positive for your model, you can disable it by passing `check_model=false` to `sample`.
+
+The keyword argument `error_on_failure=true` makes `check_model` throw an error instead of returning `false`, which is useful in scripts and tests.
+
+## Checking type stability
+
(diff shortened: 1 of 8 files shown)


shravanngoswamii:
Let me know if we should have only 1 page following structure suggested in https://github.com/TuringLang/Turing.jl/issues/2415:

> We should have a tutorial which goes through the following aspects of a model:
> 1. How to benchmark and profile a model.
> 2. How to choose the right AD backend.
> 3. How to write models in a performant manner.
>
> This is what I'm imagining (please do make suggestions!)
>
> ## 1. Benchmarking and profiling
>
> Should include:
> - Usage of TuringBenchmarking.jl.
> - Usage of a profiler, e.g. PProf.jl, both in the evaluation of the model itself _and_ in gradient computations.
>
> ## 2. Choosing AD backend
>
> This will be informed by the steps in the previous section.
>
> Also, maybe have some description (or at least point to another resource which outlines) the pros and cons of the different AD backends.
>
> ## 3. Writing performant models
>
> This comes down to a few tricks:
> - Keep LHS of `~` "simple", i.e. `x ~ filldist(Normal(), n)` instead of `x[i] ~ Normal()` inside a for-loop.
> - Use typing information, e.g. `::Type{TV}=Vector{Float64}` in the model definition, and so on.
> - Checking of type stability, e.g. using `DynamicPPL.DebugUtils.model_typed` (or maybe using Cthulu.jl?)

usage/debugging/index.qmd
```diff
@@ -0,0 +1,77 @@
...
+It complements the [Troubleshooting]({{<meta usage-troubleshooting>}}) page, which explains specific error messages: the tools here help you find problems before (or without) hitting an error.
+
+```{julia}
+using Turing
+```
+
+## Checking model validity with `check_model`
+
+`DynamicPPL.check_model` evaluates a model once and checks it for common problems, such as the same variable appearing on the left-hand side of more than one tilde statement, or `NaN` values being observed.
+It returns `true` if the check passes, and `false` (with warnings explaining the problem) otherwise.
```
sunxd3, line 23: it's worth talking a bit more on what `check_model` can and can't catch

usage/performance-tips/index.qmd
```diff
@@ -15,6 +15,51 @@ Pkg.instantiate();
...
+using DynamicPPL, LogDensityProblems, Profile
+
+model = bmodel(randn(100))
+ldf = LogDensityFunction(model, getlogjoint_internal, LinkAll(); adtype=AutoReverseDiff())
+x = rand(ldf)
+
+Profile.@profile for _ in 1:10_000
+    LogDensityProblems.logdensity_and_gradient(ldf, x)
+end
+Profile.print(; mincount=10)
```
sunxd3, line 57: is this supposed to print something? rendered doc is not showing

shravanngoswamii: It prints machine specific stuff, which can be viewed via Profilers, see: https://docs.julialang.org/en/v1/manual/profile/

Added a note below it mentioning that!

usage/debugging/index.qmd
shravanngoswamii: Can you have a look at it again and let me know if it looks good now or not?

sunxd3 (approved):

PR #703, opened 2026-09-05
Update for Turing v0.48

shravanngoswamii:
The interface guide described the AbstractMCMC v1 interface and its code cells were `eval: false`, so CI never noticed. It is rewritten on the current interface with every cell executed.

The `to_distribution` example is not executed because it needs BridgeStan and a Stan toolchain.

### DIFF
--- a/core-functionality/index.qmd
+++ b/core-functionality/index.qmd
@@ -125,13 +125,14 @@ The arguments for each sampler are:
 - SMC: resampling scheme (optional). The number of samples requested from `sample` is the number of particles in the single sweep.
 - PG: number of particles, resampling scheme (optional).
 - HMC: leapfrog step size, leapfrog step numbers.
+- MH: no arguments to propose from the prior, or a covariance matrix for a Gaussian random walk over the complete linked parameter vector, as in `MH(0.1 * I(2))`.
 - Gibbs: pairs of variable names and the component sampler for each.
 - NUTS: number of adaptation steps (optional), target accept ratio.
 
 More information about each sampler can be found in [Turing.jl's API docs](https://turinglang.org/Turing.jl).
 
 FlexiChains.jl provides plotting tools for the `VNChain` objects returned by a `sample` function.
-For more information about FlexiChains and its plotting capabilities, please see the [FlexiChains.jl documentation](https://pysm.dev/FlexiChains.jl/stable).
+For more information about FlexiChains and its plotting capabilities, please see the [FlexiChains.jl documentation](https://juliabayes.org/FlexiChains.jl/stable/).
 
 ```julia
 using FlexiChains
@@ -189,7 +190,7 @@ var_1 = mean(chn[@varname(var_1)]) # Taking the mean of a variable named var_1.
 
 The key should be a `VarName`, constructed with the `@varname` macro.
 
-FlexiChains has a very powerful indexing interface, which includes the ability to index into sub-variables: please see [the FlexiChains docs](https://pysm.dev/FlexiChains.jl/stable/turing/#Accessing-data) for more information.
+FlexiChains has a very powerful indexing interface, which includes the ability to index into sub-variables: please see [the FlexiChains docs](https://juliabayes.org/FlexiChains.jl/stable/turing#Accessing-data) for more information.
 
 ### Tilde-statement ordering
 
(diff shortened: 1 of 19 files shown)


core-functionality/index.qmd
```diff
@@ -122,16 +122,17 @@ c5 = sample(gdemo(1.5, 2), NUTS(0.65), 1000)
 
 The arguments for each sampler are:
 
-- SMC: number of particles.
-- PG: number of particles, number of iterations.
+- SMC: resampling scheme (optional). The number of samples requested from `sample` is the number of particles in the single sweep.
+- PG: number of particles, resampling scheme (optional).
 - HMC: leapfrog step size, leapfrog step numbers.
-- Gibbs: component sampler 1, component sampler 2, ...
+- MH: no arguments to propose from the prior, or a covariance matrix for a Gaussian random walk over the complete linked parameter vector, as in `MH(0.1 * I(2))`.
```
yebai:
```suggestion
- MH: propose from the prior if `MH()`, or from a Gaussian random walk if `MH(cov)` over the unconstrained parameter vector, as in `MH(0.1 * I(2))`.
```

usage/mode-estimation/index.qmd
```diff
@@ -151,6 +151,7 @@ maximum_likelihood(model; lb=lb, ub=ub)
 Turing will internally translate these bounds to unconstrained space if `link=true`; as a user you should not need to worry at all about the details of this transformation.
 
 In this case we only have one parameter, but if there are multiple parameters and you only want to constrain some of them, you can provide bounds for the parameters you want to constrain and omit the others.
+A bound must cover a whole variable: bounding only `x[1]` of a vector `x` is an error, and a bound on a name the model never reaches produces a warning rather than being ignored.
```
yebai: Bounds apply to complete tilde sites, not necessarily complete arrays. Bounding only `x[1]` works when elements use `x[i] ~ Normal()`, but fails when the whole vector uses `x ~ MvNormal(...)`.

```suggestion
A bound must cover the complete value of a tilde site.
Bounding only `x[1]` is valid when the model declares elements separately with `x[i] ~ Normal()`, but errors for `x ~ MvNormal(...)`.
A bound on a name the model never reaches produces a warning.
```

usage/vectorisation/index.qmd
```diff
@@ -213,6 +216,40 @@ LogDensityFunction(model, logdensity_function, ranges_and_transforms, sample_vec
 
 Please see the DynamicPPL API documentation for more information on these arguments.
 
+## Minibatch log densities
+
+Stochastic inference algorithms, such as those in AdvancedVI and stochastic-gradient variants of HMC, evaluate the log density on a minibatch of the data and scale the likelihood up to the full dataset.
+`subsample` builds a `LogDensityFunction` for one such minibatch.
+
+The observation has to be written with `independent_distribution`, which declares that the data are conditionally independent draws, and the full dataset has to be supplied by conditioning rather than as a model argument.
```
yebai, line 224: Minibatching requires exactly one conditioned observation, with no later probability-bearing statements. For example, placing `μ ~ Normal()` after `x ~ independent_distribution(Normal())` throws an error. The guide omits these restrictions.

```suggestion
The observation has to be written with `independent_distribution`, which declares that the data are conditionally independent draws, and the full dataset has to be supplied by conditioning rather than as a model argument.
The model must have exactly one conditioned observation, and it must be the last probability-bearing statement.
For example, placing an independent prior after that observation causes an error.
Additional likelihood contributions through `@addlogprob!` are not supported.
```

developers/inference/abstractmcmc-interface/index.qmd
```diff
@@ -14,310 +16,196 @@ Pkg.instantiate();
...
+    rng::Random.AbstractRNG,
+    model::AbstractMCMC.LogDensityModel,
+    sampler::MetropolisHastings;
+    initial_params=nothing,
     kwargs...,
 )
-    return Transition(model, spl.init_θ)
+    θ = initial_params === nothing ? rand(rng, sampler.proposal) : initial_params
+    lp = LogDensityProblems.logdensity(model.logdensity, θ)
+    return θ, MHState(θ, lp)
```
yebai, line 140: The sampler can start outside the target’s support. With seed 6 and no explicit starting point, all five returned standard deviations are negative. Require a valid starting point and check its log density.

```suggestion
    initial_params,
    kwargs...,
)
    θ = initial_params
    lp = LogDensityProblems.logdensity(model.logdensity, θ)
    isfinite(lp) || throw(ArgumentError("initial_params must have finite log density"))
    return θ, MHState(θ, lp)
```

developers/inference/abstractmcmc-interface/index.qmd
```diff
@@ -14,310 +16,196 @@ Pkg.instantiate();
...
 
-1. Make a new transition containing a proposed sample.
-2. Calculate the acceptance probability.
-3. If we accept, return the new transition, otherwise, return the old one.
-
-### Steps
-
-The `step!` function is the function that performs the bulk of your inference. In our case, we will implement two `step!` functions -- one for the very first iteration, and one for every subsequent iteration.
+The first `step` chooses the starting point.
+`sample` passes its `initial_params` keyword argument through to this method, so the sampler honours it and otherwise draws a start from the proposal distribution.
```
yebai: Update the initialisation description to match the required starting point.

```suggestion
Supply a starting point with finite log density through `initial_params`.
```

yebai:
I left a few minor comments above; otherwise, it looks good.

developers/inference/abstractmcmc-interface/index.qmd
shravanngoswamii: Applied with `initial_params=nothing` kept as the default and an explicit `ArgumentError`, because `MCMCThreads` passes `nothing` when the caller omits it and the required keyword would fail inside the destructuring instead.

shravanngoswamii:
@yebai Applied all your suggestions and few more fixes.

yebai (approved):

PR #705, opened 2026-09-19
Update for Turing v0.49

shravanngoswamii:
DynamicPPL 0.42.12 added a `multithreaded` keyword to `predict`, and it is deliberately left undocumented here.
It is defined in DynamicPPL's MCMCChains extension, but `sample` returns a FlexiChain, so `predict` dispatches into `FlexiChainsDynamicPPLExt` where the only keyword is `include_all`.
Calling it from these docs throws a `MethodError`.
Should FlexiChains gain the keyword too?

`Turing.Model`, in the variational inference tutorial, was never a real name.
`isdefined(Turing, :Model)` is `false`.


yebai:
Good work, @shravanngoswamii!

